PR #1, opened 2016-03-19
Geniverse Case 3 recreation

kswenson:
Implements both challenges from Geniverse Case 3:
1: Modify mother drake's alleles to generate a specific target offspring drake
2: Modify father drake's alleles to generate two specific target offspring drakes

Implements Case 3 as a stateful React component, leaving only a small amount of non-React scaffolding in case-3.js.


sfentress:
I think this all seems fine.

The only thing I wonder if you'd consider would be whether to cement the idea of drag and drop elements, and so to put the wrapping code (all the `DragOrganismGlowView` etc.) in the GeniBlocks views themselves.

In the alleles demo, it was really still an experiment, but if we think that this model is good, it would be great not to have to re-write it every time we had a glowing drop target.

The view could even export multiple views (`GlowView`, `GlowView.DropTarget`, or whatever), and we would ideally either package up the dnd library in the app, or degrade nicely when it didn't exist.

Do you think it's worth adding that to this PR, as this is the first case recreation with drag and drop?

sfentress:
This could also possibly wait until after PR #2, if that makes sense with the packaging.

kswenson:
I intentionally left Drag/Drop out of this round because I think it's a special case that may be handled differently and will likely be informed by the earlier efforts. Including it in GeniBlocks the way I did for React-Motion and React-SimpleTabs probably won't work because the application layer needs access to the DragDropContext, and it's not straightforward to have browserify export GeniBlocks and additional symbols (I tried). I suspect the best solution in the long run will be to have browserify build a separate geniblocks-lib which would include React, ReactDom, and BioLogica (with build flags making them optional for clients that don't need them) and then additional libraries like React-DnD that may be required by both GeniBlocks and the application layer. React-DnD would then be exposified in the geniblocks build so that GeniBlocks code wouldn't have to care where it was packaged. I actually started down that path as well but ran into some difficulties and decided to move onto other things since I'd already spent a fair amount of time on the steps to get this far. I definitely intend to get back to the question of packaging React-DnD properly when the opportunity presents itself.

examples/case-3/case-3.js
```diff
@@ -0,0 +1,483 @@
...
+  state = {
+    selectedIndex: null
+  }
+
+  constructor(props) {
+    super(props);
+    document.addEventListener('keydown', () => enableButtonFocusHighlight());
+  }
+
+  breed = () => {
```
pjanik: I don't really know this project, but accidentally noticed this PR. 
Out of curiosity, I guess `breed = () => {` isn't just a fancy way to say `breed() {`, but it provides auto-binding right? Nice.

sfentress: Actually, I don't really understand it myself. You can use `this` in the same way when you use `breed() {`, and playing with the code, it all seems to work fine when re-writing all of the `breed = () => {`-style functions as `breed() {` instead.

So is there a reason to use `breed = () => {`?

pjanik: Not sure, that's why I was asking. :wink: But in general an arrow function `() => { ... }` is equal to `function () { ... }.bind(this)`. But I've never seen it in the class declaration. So, I was wondering if that works in this case and e.g. you can just call `setTimeout(this.breed, 1)` instead of `setTimeout(this.breed.bind(this), 1)`.

kswenson: I was following the example at [React on ES6+](https://babeljs.io/blog/2015/06/07/react-on-es6-plus#arrow-functions) and overgeneralized to that pattern for all methods. It's only useful on methods that are to be used as callbacks and the utility (as Pyotr suggested) is that it eliminates the need to bind at point of use. I've updated the PR accordingly to only use it where appropriate and simplified the corresponding call sites.

sfentress:
This looks fine to me. I'd like to get back to the discussion of providing pre-wrapped DnD components with our views at some point, but it doesn't need to be now.

PR #2, opened 2016-03-23
Improvements to build and module reference conventions

kswenson:
Improvements to build and module reference conventions for GeniBlocks:
- Rename app.js/app.css to geniblocks.js/geniblocks.css.
- React-Motion and React-SimpleTabs built into GeniBlocks library and required by components that need it.
- React exposified; GeniBlocks components require it explicitly rather than assuming a global.

Note: Currently depends on [PR#1](https://github.com/concord-consortium/geniblocks/pull/1). I'll rebase once that code is reviewed.


gulp/tasks/default.js
```diff
@@ -2,13 +2,13 @@ var gulp = require('gulp');
...
-    gulp.watch(config.browserify.watch,   ['browserify-app']);
+    gulp.watch(config.browserify.watch,   ['browserify']);
     gulp.watch(config.css.watch,          ['css']);
     gulp.watch(config.resources.watch,    ['resources']);
     gulp.watch(config.examples.watch,     ['examples']);
     gulp.watch(config.examplesJS.watch,   ['examplesJS']);
 });
 
-gulp.task('build-all', ['browserify-app', 'browserify-minify-app', 'vendor', 'css', 'resources', 'examples', 'examplesJS']);
+gulp.task('build-all', ['browserify', 'vendor', 'css', 'resources', 'examples', 'examplesJS']);
```
sfentress: Why was `browserify-minify-app` removed as a task? The task is still defined now in `browserify-task.js`, but is never run.

If we want to distinguish production builds from dev builds, we could do that, but note that it seems we are still building both css files in the `build-all` task.

package.json
```diff
@@ -26,6 +33,7 @@
     "gulp-streamify": "^1.0.2",
     "gulp-stylus": "^2.1.0",
     "gulp-uglify": "^1.5.1",
+    "react": "^0.14.7",
```
sfentress: Are we using the React brought in by the `devDependencies` at all? We're still relying on the bower version, and the new exposify addition doesn't seem to depend on this being listed in devDependencies at all.

I'm also confused by the addition of React in `peerDependencies`. This just means that consumers who pull in Geniblocks via npm will also pull in React, but at the moment, at least, we're agnostic about how the consumer pulls in React -- they could be using bower, like we do for our examples, or something else.

(Ultimately my guess is that this will all start becoming clearer once we create the Geniverse 2.0 application, which is something I feel like we should be getting to soon.)

gulp/tasks/default.js
kswenson: Good catch. My goal was to simplify and make it more consistent with the way the CSS is handled (and the way browserify is handled in CFM), but I failed to add 'browserify-minify-app' to the 'browserify' task in browserify-task.js. I'll make that change as well as putting back the 'browserify-app' in watch so we don't minify in that case.

sfentress:
It would probably make sense to remove f9a5ae3 from this PR, so it doesn't cause conflicts.

package.json
kswenson: As near as I can tell, this is the model that React-based libraries seem to be settling on -- declare a peer-dependency to acknowledge the run-time requirement and a dev-dependency to meet the requirement during development. React-DnD, React-Motion, and React-SimpleTabs all follow that model. Now that React-Motion and React-SimpleTabs are dependencies of GeniBlocks, if we don't have the dev-dependency, then on `npm install` we get
```
npm WARN EPEERINVALID react-simpletabs@0.7.0 requires a peer of react@0.12.x - 1.x but none was installed.
npm WARN EPEERINVALID react-motion@0.4.2 requires a peer of react@>=0.13.2 || ^0.14 but none was installed.
```
Peer dependencies don't cause React to be pulled into the bundle, however, so at runtime it is still the case that React is provided by the caller. The difference now is that the GeniBlocks code can/must `require` or `import` React explicitly rather than assuming that it's available as a global and `exposify` is used to make the global React available via `require` or `import`, thus decoupling the GeniBlocks code from the way in which React is provided for it. Moving forward, React will be a more explicit dev-dependency in the upcoming unit test harness PR.

kswenson:
Rebased and squashed for merging.

package.json
sfentress: > Now that React-Motion and React-SimpleTabs are dependencies of GeniBlocks, if we don't have the dev-dependency, then on npm install we get

Hmmm... but it still has exactly zero effect on development or building, unless I'm mistaken. We never touch the package at all. It seems to me like if we're adding the package simply to remove the npm warning, that signifies we're doing something wrong.

I think if we bring react in as a package, we should work out how to use it as a package, and not use the bower version. Since we're already building the examples, it seems like this wouldn't be too hard. It doesn't look like we're currently using Browserify on the examples, though, so a simple `import` statement won't work until we change how we're building them.

> The difference now is that the GeniBlocks code can/must require or import React explicitly

This doesn't seem related to the listing of React as a peerDependency or as a devDependency, since we can do this even if we don't list it at all.

> As near as I can tell, this is the model that React-based libraries seem to be settling on

Having read through a bunch of this discussion online, I agree with you on this. And now that I understand that peerDependencies are not brought in (which seems counter to this statement in https://nodejs.org/en/blog/npm/peer-dependencies/ that "Now, when installing chai-as-promised, the chai package will come along with it," but no matter because that's obviously incorrect), I don't have a problem with listing it as a peerDependency.

So... I guess in the end my only thought would be to suggest that we use our npm version of React in our demos, and get rid of the bower version. I see https://github.com/jxnblk/rebass does this as well. It might be a good chance to explore how a consumer (us, pretty soon) really *should* be using this package.

kswenson: For now the dev-dependency is following convention and laying the groundwork for the unit testing work I'm doing, which will require it. I completely agree that we should figure out the right way for our demos and applications to consume GeniBlocks along with the other required libraries (React, ReactDOM, Biologica, ...), but I don't think it needs to be part of this PR.

PR #3, opened 2016-04-13
Unit test harness and initial GeniBlocks component unit tests

kswenson:
- Unit test harness based on enzyme, chai, shallow rendering, jsdom
- Accommodate BioLogica.js changes for ease of testing
- Fix issues found during testing (e.g. classNames -> className)
- Refactor some components for ease of testing (e.g. StatsView)
- Add filterAlleles() and computeTraitCountsForOrganisms() to GeneticsUtils
- Convert some constants to be properties passed by caller

### DIFF
--- a/src/code/components/allele-filters.js
+++ b/src/code/components/allele-filters.js
@@ -1,4 +1,4 @@
-import {PropTypes} from 'react';
+import React, {PropTypes} from 'react';
 
 const AlleleFiltersView = ({species, hiddenAlleles=[], disabledAlleles = [], onFilterChange}) => {
   let hiddenGenes = new Set,
--- a/examples/basic-examples/genome.html
+++ b/examples/basic-examples/genome.html
@@ -21,7 +21,7 @@
           React.createElement(GeniBlocks.GenomeView, {
             org: dragon,
             alleleChanged: function(chrom, side, prevAllele, newAllele) {
-            dragon.genetics.genotype.chromosomes[chrom][side].alleles.replaceFirst(prevAllele, newAllele);
+            dragon.genetics.genotype.replaceAlleleChromName(chrom, side, prevAllele, newAllele);
             dragon = new BioLogica.Organism(BioLogica.Species.Drake, dragon.getAlleleString(), 1);
             render();
           }}),
--- a/examples/case-1/challenges.js
+++ b/examples/case-1/challenges.js
@@ -124,7 +124,7 @@ function render() {
       hiddenAlleles: hiddenAlleles,
       style: {marginTop: 50, top: 50},
       alleleChanged: function(chrom, side, prevAllele, newAllele) {
-        yourDrake.genetics.genotype.chromosomes[chrom][side].alleles.replaceFirst(prevAllele, newAllele);
+        yourDrake.genetics.genotype.replaceAlleleChromName(chrom, side, prevAllele, newAllele);
         yourDrake = new BioLogica.Organism(BioLogica.Species.Drake,
                                             yourDrake.getAlleleString(),
                                             sexOfYourDrake);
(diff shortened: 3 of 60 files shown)


src/code/components/allele-filters.js
```diff
@@ -1,4 +1,4 @@
-import {PropTypes} from 'react';
+import React, {PropTypes} from 'react';
```
sfentress: What's the purpose of this when we've turned off the `react-in-jsx-scope` rule? If this is preparing for a time when `React` might not be a global, should we turn back on the rule?

kswenson: Unit tests fail without the import of React, so at this point we could enable the `react-in-jsx-scope` rule if we were so inclined or figure out a way to load React globally in setup.js, analogous to the way BioLogica is handled now.

sfentress:
I think all this is fine, and we can pull it in, but I think it may warrant a (possibly future) conversation about what the purpose of React view tests are.

It looks like 99% of the component tests here are essentially:

MyComponent.js:
```

<div className="foo">
  <div className="bar">
    [...]
  </div>
</div>
```

MyComponentTest.js
```
assert(x.find('div').hasClass('foo'));
assert(y.find('div').hasClass('bar'));
```

i.e. that we find all the elements we specified in the jsx.

But all this seems to be doing is specifying what elements should be there, twice. If we make a change to a view, we have to change in once in the component, and once in the test. 

In these kinds of tests, what kinds of situations do we expect the tests to fail? Only when we deliberately change the view itself.

I guess the tests that I could see being "fail-able" in more interesting ways would be, e.g. Does the `organism` component create an img tag with the *correct* url? Does the `chromosome` component show all the pulldowns with the correct state? If you modify a pulldown, do we get the correct callback? Does the `genome` view show all the correct chromosomes, and correctly pass through the event when you modify a pulldown?

It's reasonable for these current tests to be the basis of more in-depth testing, but even then I worry about the busy-work these represent: is the expectation that every component has tests that assert the presence of every `div` and `span`, and every time we change the components are we expected to update the tests?

kswenson:
I think 99% is a little high ;-) -- I was thinking about that question and have a couple of thoughts. First, a common type of failure is simply that the component doesn't load at all due to syntax error or undefined dependency. Having at least a simple existence test that something reasonable got rendered provides value in detecting those situations. If we've used some construct that blows up on IE, for instance, I would expect these tests to catch that. Some of the React testing libraries out there support/recommend component tree comparison, i.e. the rendered view hierarchy should look exactly like _this_. I didn't want to go down that route. My goal was to find a balance between verifying that the component did its job without writing a test that would be invalidated by changes to internal implementation. So, OrganismView is expected to render an  tag, but the test doesn't rely on precisely where in the hierarchy that  tag exists. If no  tag gets rendered that's likely a problem and if we ever change the implementation of OrganismView so that it doesn't render an  tag that seems like a reasonable time to revisit the unit tests. In views that support allele filtering, I figured it made sense to test that the number of rendered children was consistent with filtering having been done correctly. There's always a question of diminishing returns with unit tests in general, and view unit tests in particular. In what I did for this PR I tried to balance that by testing where it seemed there was a plausible point of failure in the component's contract with its client without writing tests that depended too heavily on (and hence would be invalidated by changes to) implementation details of the component.

For your specific examples, `OrganismView` relies on BioLogica for its mapping from Organism to URL, and BioLogica has an extensive set of tests for it, so I didn't want to be redundant with those, although I wouldn't object to adding a spot-check test to what's already there. I did add some tests for callbacks in some of the components but admittedly didn't exhaustively add such tests. In some respects the shallow rendering approach isn't ideally suited for that (you have to specify the event manually in the test which potentially just propagates any misconceptions one might have about the contents of the event) and so there may be another approach to consider there.

sfentress:
Ok. I think it would be good to keep thinking more about testing the callbacks, and also the minimum necessary test to ensure that a component is doing the correct thing (e.g. only testing the `` tag in `OrganismView` instead of also the wrappers), and performing the interacting with biologica correctly (e.g. the url in `OrganismView`, to confirm the view knows how to create an Organism correctly -- which is the most complex thing the view is doing, and can easily be broken by, e.g. importing a newer version of biologica), but this seems fine otherwise.

👍

PR #6, opened 2016-04-15
Mitigate case 3 browser drag/drop inconsistencies

kswenson:
- Dragging works on Firefox (but must select first)
- DragImage is the drake on all tested browsers
- Click selection/deselection works as expected
- Mitigates [#117663369]
- Fixes [#117663015]


kswenson:
Given that Sam is likely to be unavailable for a while I'm going to go ahead and merge this.

PR #19, opened 2016-06-08
Challenge award display

dstrawberrygirl:
Integrated into ModalAlert, images are overlaid one on top of another so the final representation could be a series of stars, pieces of a coin, jigsaw, etc. Includes sample images for testing, the award display is integrated into the final "challenge complete" part of the GV2 case 1 example and a simulated "completed cases 1-5" dialog result shown on successful completion of the case 5 example.


src/code/components/challenge-award.js
```diff
@@ -0,0 +1,63 @@
...
+    let challengeId = 0, progress = [];
+    if (this.props.challengeProgress){
+     challengeId = this.props.challengeProgress.id;
+     progress = this.props.challengeProgress.progress; //challengesJson.Challenges[this.props.challengeId].Progress;
+    } else return null;
+    
+    if (challengeId === 0 || !progress || progress === []) 
+      return null;
+    
+    let baseUrl = `http://127.0.0.1:8080/gv2/images/challenge${challengeId}`;
```
sfentress: Wouldn't this work as a relative url?

src/code/components/modal-alert.js
```diff
@@ -77,6 +80,7 @@ class ModalAlert extends React.Component {
               onHide={this.props.onHide} >
         <div style={dialogStyle()} >
           <h4 id='modal-label'>{this.props.message}</h4>
+          <AwardView challengeProgress={this.props.challengeProgress} />
```
sfentress: This works, but is a little confusing to me because it looks like an AwardView is always expected to be rendered, but then AwardView itself might return false.

Another way could be 

    if (this.props.challengeProgress) {
        <AwardView challengeProgress={this.props.challengeProgress} />
    }

I don't know if one is a more idiomatic React convention than the other, so I'm also fine leaving it as is.

sfentress:
:+1:

PR #21, opened 2016-06-09
Geniverse single project

sfentress:
This PR is my second attempt at creating a "Geniverse-2.0" project (see #20 
The two) while still maintaining a separate, independent library of React components that could be used by multiple genetics projects.

This version keeps a standard structure for a typical React application codebase, with the views in the components folder. However, these views are also built to /dist as a separate library, and this library is also what is used exclusively in the /examples folder, which can act as confirmation that the library works independently.

The GeniBlocks library is basically untouched, and is not bundled with React, so React must be brought in seperately by the applications that use it. The Geniverse project builds React into the application.

Currently the Geniverse-2.0 application just builds a Hello World app, which can be found at public/gv2.


sfentress:
@dstrawberrygirl If you have a moment to look at this (and the other PR too, if you want), I'd love feedback.

(@kswenson you too, but you're supposed to be on vacation, so close the computer.)

PR #24, opened 2016-07-05
Challenge award coins

dstrawberrygirl:
Remove calculated score and grab scores from challengeProgress in state
Remove placeholder images and replace with much-improved coin sprite tiles
Display coin parts on successful completion of a challenge
Choose appropriate coin piece based on number of pieces in the coin and the current piece number


src/code/reducers/challengeProgress.js
```diff
@@ -36,17 +36,17 @@ export function getChallengeName(case_, challenge, trial){
...
+//     } else {
+//       challengeAwardProgress.push(-1);
+//     }
+//   }
   
-  return challengeAwardProgress;
-}
\ No newline at end of file
+//   return challengeAwardProgress;
+// }
```
sfentress: Do you want to just delete this all out? If there's stuff that needs to be resurrected, it can always be found in the git history.

sfentress: D'oh, you already did this in the next commit. Hadn't seen that yet.

src/stylus/coins.styl
```diff
@@ -0,0 +1,54 @@
...
+  position: absolute;
+  height: h;
+  width: w;
+  background-image: url("../resources/images/coin-background.png");
+  &.piece
+    for pieces in 2..5
+      // coins in 2 3 4 or 5 pieces
+      for pieceNum in 1..pieces
+        // for each piece in pieces
+        &.pieces{pieces}
```
sfentress: Would it make sense to add

`background-image: url("../resources/images/coin-pieces-(pieces).png");`

right here, and then you wouldn't have to also add the background image-name calculation to challenge-award.js? It seems like it would be nice to have it all defined in the CSS, instead of partly in the CSS and partly in the JS (especially as it hard-codes this specific image set in that component).

PR #28, opened 2016-07-13
Add translations, authored ITS messages

sfentress:


sfentress:
Hi @dstrawberrygirl, if you could take a look at this when you have a moment, that would be great. Thanks!

dstrawberrygirl:
Looks good @sfentress - there are a couple of whitespace changes in the egg-game template that I think would be best fixed separately, otherwise 👍

sfentress:
You're saying it would be better to take the whitespace changes out of this PR, and put them in a different commit?

(To be clear, these changes are whitespace removals -- sometimes it looks the other way around on GitHub.)

src/code/utilities/translate.js
```diff
@@ -0,0 +1,29 @@
+import translations from './lang';
+
+const defaultLang = "en_us",
+      varRegExp = /\$\{\s*(\d+)\s*\}/g,
+      error = "** TRANSLATION ERROR **";
+
+const translateString = (key, lang) => translations[lang][key] || key;
```
kswenson: Perhaps a test for existence of translations[lang] in case of request for language for which there are no translations.

kswenson:
@sfentress I took a quick look as well. Looks like a nice addition and I like the parameterized strings! I made one line comment on the translateString() function.
👍

sfentress:
Thanks @kswenson, sounds reasonable. Added a test and fix in 51dc154, continuing with the idea of showing raw untranslated strings to the user if there is a translation failure.

(This may all change, BTW, as the GUIDE folks are proposing changes to the translation design.)

sfentress:
@dstrawberrygirl This has been rebased to remove the changes to `egg-game.js`. Do you think it's ok to merge?

PR #30, opened 2016-07-21
Refactor reducers

sfentress:
This splits up the main reducer file into multiple reducers.

It also re-writes many of the actions that produced side-effects that weren't obvious from the action name -- e.g. submitting a drake would also pop-up a message. We now have dedicated reducers for things like messages, and many of the actionCreators now dispatch multiple actions, with logic inside the action creators themselves.


src/code/actions.js
```diff
@@ -275,9 +363,33 @@ export function initiateDelayedFertilization(delay, gamete1, gamete2) {
   };
 }
 
-export function keepOffspring() {
+export function _keepOffspring(index, success) {
```
knowuh: FWIW: When I see a function name starting with an underscore `_`, I assume it isn't an externally available function, but `export` proves me wrong here.

sfentress: You're right, thanks @knowuh. 

In theory they could also be public, but we don't yet have a good naming distinction between actions that just do a tiny piece of a larger action, and the larger actions that you would actually want to call.

Probably the two "_littleActions" I just un-exported now in 00a8935 could not be action-creator functions at all, and instead the bigger actions could just dispatch the action object itself. Not sure -- we'll probably continue to change conventions as we go on.

PR #33, opened 2016-09-15
Project maintenance/preparation

kswenson:
- Fix demos/examples
- Fix unit tests
- Fix eslint warnings/errors
- Update project dependencies
- Addresses [[#130463499]](https://www.pivotaltracker.com/story/show/130463499)


dstrawberrygirl:
I think these all look good - the only question I have relates to switching from string representations "male"/"female" to BioLogica.MALE/BioLogica.FEMALE when it comes to the ITS integration. I can't remember the final discussions on this, @sfentress may recall this better than I since he looked at the ITS part in more depth.

sfentress:
Pretty sure ITS wanted to receive messages that included "male"/"female" over 0/1, but I don't think that's affected by this PR.

kswenson:
If memory serves, I originally had the buttons using strings. At some point they were changed to use the BioLogica constants. In this PR I'm just updating the unit test to correspond to the current state of the code. What we send ITS can certainly be independent of what the buttons use internally.

PR #35, opened 2016-09-28
Egg Game II

kswenson:
- author can specify challenge type ("create-unique" or "match-target")
- author can specify whether user drake is hatched/shown automatically
- target drake is shown (and pen view is hidden) in target matching challenges [#129879003]
- user can fertilize once gametes are complete [#129879183][#129879515]
- egg hatches (Egg Game 2a) or not (Egg Game 2b) [#129879183][#129879515]
- after fertilization user can submit drake or reset gametes [#129879109]
- dialog gives feedback on whether submitted drake matches [#129879109]
- drake is visible while feedback dialog is displayed
- Egg Game 2a (hatched drake) and 2b (unhatched egg) each contain three trials
- gold, silver, or bronze coin piece provided upon completion of challenges [#120794161]

With 20/20 hindsight, it might have made sense to split this up over a couple of PRs, but I kept thinking that just one more story wouldn't make that much difference.

### DIFF
--- a/src/resources/authoring/gv2.js
+++ b/src/resources/authoring/gv2.js
@@ -86,7 +86,9 @@ window.GV2Authoring = [
       },
       "targetDrakes": [{},{},{}],
       "hiddenAlleles": "t,tk,h,c,a,b,d,bog,rh"
-    },
+    }
+  ],
+  [
     {
       "template": "GenomePlayground",
       "initialDrake": {
@@ -94,6 +96,6 @@ window.GV2Authoring = [
         "sex": 1
       },
       "hiddenAlleles": "t,tk,h,c,a,b,d,bog,rh"
-    }
+    },
   ]
 ];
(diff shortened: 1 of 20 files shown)


src/resources/authoring/gv2.js
```diff
@@ -86,14 +86,16 @@ window.GV2Authoring = [
...
+  [
     {
       "template": "GenomePlayground",
       "initialDrake": {
         "alleles": "a:T,b:T,a:w,b:w,a:h,b:h,a:A1,b:A1,a:C,b:C,a:B,b:B,a:D,b:D,a:rh,b:rh,a:Bog,b:Bog",
         "sex": 1
       },
       "hiddenAlleles": "t,tk,h,c,a,b,d,bog,rh"
-    }
+    },
```
sfentress: Will some browsers throw an error on this comma?

kswenson: 1. I don't think any browsers we care about would notice. (As I recall, IE8 was the last browser to have that problem.)
2. Even so, I generally try to avoid it as a matter of general code hygiene, so good catch.
3. ESLint must have a setting to catch extraneous commas.
4. It's already been changed in subsequent PRs to go on to Egg Game III rather than circling back to the Genome Playground.

PR #36, opened 2016-10-07
Egg Game III: Inspectable Eggs

kswenson:
Attempting to implement Egg Game III in manageable chunks, the first chunk implements a set of eggs that are inspectable. Clicking on a basket indicates whether the selected egg belongs in that basket. I've already assigned @sfentress to review the large Egg Game II PR, so I'm assigning this to @dstrawberrygirl for equitability. Feel free to reassign if the two of you think there's a better way to handle it. The code is testable at http://concord-consortium.github.io/geniblocks/gv2/#/3/1. There's a lot more work to be done and the goal is to have something to put in front of students next week, so expect a spate of PRs in the coming days.

Note that this PR includes the Egg Game II PR. Once the Egg Game II PR is merged, I'll rebase this one to eliminate the overlap.


src/code/components/basket-set.js
```diff
@@ -0,0 +1,64 @@
+import React, {PropTypes} from 'react';
+
+/**
+ * @param {number} rows - Option number of rows. If defined, it will be fixed at that. Otherwise, it
+ *                        will default to 1 when there are no orgs, and grows as more rows are needed.
+ * @param {number} tightenRows - If given, will shrink the vertical height of the pen by this amount
```
dstrawberrygirl: Do you need to specify this manually, or could it be derived from inspecting the overall screen size and content bounds?

kswenson: It's just left-over copy-paste. I actually eliminated all sizing calculations from the somewhat analogous BasketSetView and EggClutchView, leaving all the layout in CSS. Clearly, I need to fix the comments, however. 😉

test/actions/navigate-to-challenge.js
```diff
@@ -47,19 +45,14 @@ describe('navigateToChallenge action', () => {
 
       let nextState = reducer(initialState, navigateToChallenge(1, 2));
 
-      expect(nextState).toEqual(initialState.merge({
+      // by starting with nextState and merging in what we expect,
+      // we eliminate all other properties from consideration
+      expect(nextState).toEqual(nextState.merge({
```
sfentress: I'm not sure about this. It seems like an easy way to have unintentional changes to the state be missed by the test. If the reducer changes a different property other than the ones listed here, the test will still pass.

I think it's ok to have explicit properties that we know will change and we don't care about their values (in this case we know that `challengeProgress` etc will change due to the navigation), but we don't want to miss other state changes.

kswenson: Good point, but the way it was before I had to fix the test after every unrelated state change, which didn't seem ideal, either. I'll see if I can come up with a better solution.

kswenson:
@dstrawberrygirl 
Layout Issue -- For this preview I hard-coded an absolute layout to get something working for the demo meeting. I need to go back and flex the high-level layout as well.
Game issue -- Right now the code is just generating eight eggs at random, so there's no guarantee that every basket will get an egg. If we want to guarantee a particular distribution, we'll need to use something like the TrialGenerator to randomize eggs with constraints.

sfentress:
Looking over Egg Game 3, I think my main comment is that it's using the template's internal state in cases where it can be using the Redux state. Clicking on an egg does not result in an action that changes the game's underlying statet. This means that the state isn't savable and the actions don't go through the middleware (so we don't get logging, ITS integration, etc).

In the other templates, `setState` is only used for transient, non-game-state states, specifically animation. Here `setState` is used for the egg selection, as well as the currently-selected basket, and even the parents (the last not being that relevant, but in the other templates, all the visible drakes are bound directly to the underlying state, but here the binding is lost after mounting).

I think as a general rule, anything that we can consider "undoable" or "loggable" should be an Action that updates the underlying state through a reducer (whether or not we ever implement "undo").

It's true that we may want to refactor how we divide up actions, and pass them to the templates, as we currently have these very long lists of actions, but that's a separate issue.

kswenson:
@sfentress I completely agree about the use of state. It wasn't immediately clear to me how selection should be represented in the Redux state, so I figured I'd start by getting it to work in local state and then figure out how to represent it in Redux state with actions and reducers as a subsequent step. I think that gets us into some interesting architectural questions about "global" state, i.e. state that is potentially common to all templates, versus template-specific state. I think it would be useful to have a larger conversation about what the state should look like and how best to extend it to support a set of arbitrary yet-to-be designed additional templates.

kswenson:
@sfentress @dstrawberrygirl Thanks to both of you for reviewing the code and making comments. You managed to zero in on several areas where I took the expedient approach to solving a problem for this first iteration with the understanding that there would be subsequent iterations to improve things. There's value in getting feedback on early iterations that I missed by not bringing you in to review Egg Game II earlier in the process. For the sake of visibility, I will create PT stories for some of the issues raised in the review process. Thanks again.

kswenson:
The [Egg Game II PR](https://github.com/concord-consortium/geniblocks/pull/35) has been approved and merged, so I've rebased this branch to eliminate the overlapping commits. I've created [#131899679](https://www.pivotaltracker.com/story/show/131899679), [#131901473](https://www.pivotaltracker.com/story/show/131901473), [#131900015](https://www.pivotaltracker.com/story/show/131900015), and [#131900621](https://www.pivotaltracker.com/story/show/131900621) to capture the issues raised on review so far. @sfentress @dstrawberrygirl Have you completed your reviews?

kswenson:
It comes back to consideration of "global" (cross-template) state and template or challenge state. There are global actions (e.g. GAMETES_RESET) which hard-code the assumption that the drake at index 2 is special. In the previous two templates, 0 and 1 refer to parent drakes, 2 to the user-created drake, and 3+ to template-specific drakes. In Egg Game III, I don't have a user-created drake so it seemed to make sense to leave it blank.

PR #38, opened 2016-10-14
Egg Game III: animation and interaction

kswenson:
- incorporate new basket un/selected images
- selecting an egg selects/highlights all baskets [#131750701]
- selecting a basket selects that basket and deselects other baskets [#131750685]
- selected egg animates to basket after clicking basket [#131751663]
- eggs hatch after moving to baskets [#131750759]
- user receives feedback on whether correct basket was chosen [#131752557]
- egg selection indicated by opacity rather than image change [#132180671]
- after correct basket selection, drake simply fades away [#132183603]
- after incorrect basket selection, egg returns to original location and baskets are highlighted [#132183603]

Support url parameters for testing out some alternative UI options
- 'hatchInPlace' option "hatches" drakes "within" their egg rather than moving to the basket to "hatch"
- 'collectInBasket' option collects eggs in baskets after correct selection rather than fading them out


kswenson:
@sfentress @dstrawberrygirl Another fairly large PR which you can decide between you who should review. The goal for this week was to get to a point at which the animation and interaction could be evaluated internally, so I haven't addressed the under-the-hood issues that were raised in the review of the previous PR. There are PT stories for them, however, and now that the basic interaction is available for feedback and the classroom test is still a couple weeks out I expect to be addressing some of those issues shortly.

sfentress:
I'm happy to look at this this morning.

But just to check on the intent of the PR -- is there a need to rush a merge of the PR, if you're already deploying testable builds? That is, what's the advantage of accepting PRs and spinning off PT stories based on the issues, rather than leaving the PR open?

dstrawberrygirl:
A general comment, and since I can't find a PT story for it - a lingering issue that goes back to when I was working on this, resizing the window mid-game means the animation target calculations will break and eggs will move to the wrong destination. Ideally, we need a handler for window resize events and some way to update animation position values for all the games that use these sorts of animations.

src/code/actions.js
```diff
@@ -268,12 +271,42 @@ export function submitDrake(correctPhenotype, submittedPhenotype, correct, incor
...
+}
+
+export function acceptEggInBasket() {
+  return (dispatch, getState) => {
+    const { submittedEggIndex, submittedBasketIndex } = getState();
+    dispatch(_acceptEggInBasket(submittedEggIndex, submittedBasketIndex));
+  };
+}
+
+function _submitEggForBasket(eggIndex, basketIndex, isCorrect) {
```
sfentress: Here, and in `acceptEggInBasket`, `rejectEggInBasket`, and in the final state of the `baskets` array updated by the reducer, we only have a reference to an `eggIndex`. Thinking both to our ITS integration and the Paul Horwitz of the future trying to interpret logs, this index is fairly opaque because it doesn't directly refer to anything in the state object. I think in all other cases where we've passed around an index in an action, or referred to an index elsewhere in the state, we've explicitly referred to something you can look up in the state, like `drakes[i]`. (This makes these indexes rather like using normalized JSON)

A couple ways this could be more explicit: (1) pass the drake index directly, instead of the eggs index -- `eggs[0]` is basically `drakes[3]`, it seems; (2) for aid in ITS/research, passing the genotype of the drake, like we do in `submitDrake`.

src/code/reducers/index.js
```diff
@@ -73,6 +75,33 @@ export default function reducer(state, action) {
...
+        submittedEggIndex: null,
+        submittedBasketIndex: null
+      });
+    }
+    case actionTypes.EGG_ACCEPTED: {
+      return state.merge({
+        submittedEggIndex: null,
+        submittedBasketIndex: null
+      });
+    }
```
sfentress: It seems that this properties only exist so that we can hold onto a reference of which egg and basket were selected between the `submitEggForBasket` and the `accept/rejectEggFromBasket` actions, which have a modal dialog between them. The fact that these properties only exist between these times makes it seem like a work-around for the fact that we can't easily pass properties to the modal dialog.

It seems like updating the `MODAL_DIALOG_SHOWN` action to support arguments, and then the ModalDialogContainer could use those arguments when dispatching the action, would be a more generally-useful solution.

src/code/reducers/index.js
```diff
@@ -73,6 +75,33 @@ export default function reducer(state, action) {
         challengeProgress: progress
       });
     }
+    case actionTypes.EGG_SELECTED: {
+      return state.merge({
+        selectedEggIndex: action.eggIndex
+      });
+    }
```
sfentress: Rather than having these selected indexes at the top level, why not add `selected` properties to members of the `drakes` array, and eventually the `basket` array (once that selection action is Redux-ified)? This would seem to be more generalizable to other activities where a user is selecting drakes, which will happen often, if we add the rest of the Geniverse activities. Selecting an egg could then call a totally-generic `SELECT_DRAKE` action.

Not only would this save us needing to put these indexes at the top level, it would be easier for your views. Instead of needing to pass around indexes, the object representing a `basket` simply has a property for whether it's selected, which seems more React-y.

(It adds a couple lines of code when you need to de-select all other items, but removes lines of code elsewhere.)

Of course, could be added to a template-specific section, but I don't think they need to be, since selecting drakes and other members of arrays is something we'll need to do again and again.

sfentress:
It would also be nice to have tests of the new actions.

src/code/templates/egg-sort-game.js
```diff
@@ -29,6 +29,13 @@ export default class EggSortGame extends Component {
     onEggPlaced: PropTypes.func.isRequired
   }
 
+  state = {
+    selectedBaskets: [],
+    clickedBasket: null,
+    selectedEggIndex: null,
+    selectedEgg: null
+  }
+
```
kswenson: I believe it's [the ES6+ way](https://babeljs.io/blog/2015/06/07/react-on-es6-plus#property-initializers).

kswenson:
@dstrawberrygirl I added [#132334089](https://www.pivotaltracker.com/story/show/132334089) for the window resize bug.

src/code/templates/egg-sort-game.js
kswenson: The link you gave describes initializing state in the constructor as the ES6 way and using property initializers as the ES7 way.

sfentress: Ah, indeed. :+1:

src/code/actions.js
```diff
@@ -268,12 +271,42 @@ export function submitDrake(correctPhenotype, submittedPhenotype, correct, incor
...
   let incrementMoves = !isCorrect;
   return{
-    type: actionTypes.EGG_PLACED,
```
kswenson: Done

src/code/actions.js
kswenson: Suggestion #2 would have to deal with the issue that there can be multiple eggs/drakes with the same genotype/phenotype, so I went with #1 and renamed the field to `eggDrakeIndex`.

src/code/reducers/index.js
kswenson: Done

src/code/reducers/index.js
kswenson: Done. I like the way the passing arguments to actions through the modal dialog turned out. The other use of the `submittedEggIndex` and `submittedBasketIndex` properties was to trigger post-dialog animations, which is now handled through more generally useful `correct` and `errors` counters.

kswenson:
> what's the advantage of accepting PRs and spinning off PT stories based on the issues, rather than leaving the PR open?

@sfentress There's a larger conversation to be had here, perhaps in a retrospective, but my thinking is that PT stories are how we communicate progress to PIs and provide a means for them to prioritize the work; creating PRs is how we `Start` and `Finish` PT stories; and reviewing PRs is part of validating the work for `Delivery`. As such, the improvements that are suggested on code review should be written up as stories for prioritization, unless the problems are sufficiently serious that they should prevent `Delivery`. If those code review stories are deemed lower priority than other feature stories -- as has been the case here with an upcoming classroom test giving user-visible features higher priority than under-the-hood improvements -- and we require that the PR must remain open until all review issues are addressed, then PRs end up being open indefinitely and one must either juggle multiple open PRs or put everything into one omnibus "Egg Game III" PR. Furthermore, if PRs are open for long periods of time without review approval, it becomes unclear when/whether the corresponding PT stories can be `Delivered`, which then muddies communication of progress. It seems more useful to me to have smaller PRs targeting smaller pieces of functionality. Thus, if the essential goal of a PR has been met and the issues that arise on code review are amenable to being captured as PT stories and addressed in a subsequent PR, the PR can be approved once the appropriate PT stories have been created. Of course, if there are serious bugs in the implementation of the PR and/or the implementation is fundamentally flawed in some way then it is reasonable to withhold approval until those fundamental issues have been addressed.

kswenson:
@sfentress @dstrawberrygirl I've addressed most of the issues raised on review and created PT stories for a couple that have been prioritized below other stories. I've also addressed a number of issues from Frieda based on feedback from the earlier PR. Pending your approval, I'd like to merge this PR and `Deliver` the corresponding PT stories.

kswenson:
> It would also be nice to have tests of the new actions.

Agreed. I decided to create a PT story in this case. I intend to do a round of testing improvements once I finish with the basic Egg Game III implementation, which should be in a couple of days.

src/code/actions.js
```diff
@@ -295,25 +295,23 @@ function _rejectEggFromBasket(eggIndex, basketIndex) {
...
 }
 
-export function rejectEggFromBasket() {
-  return (dispatch, getState) => {
-    const { submittedEggIndex, submittedBasketIndex } = getState();
-    dispatch(_rejectEggFromBasket(submittedEggIndex, submittedBasketIndex));
+export function rejectEggFromBasket(args) {
+  return (dispatch) => {
+    dispatch(_rejectEggFromBasket(args.eggDrakeIndex, args.basketIndex));
   };
```
sfentress: Small detail: I think there's no need to use redux-thunk here to call a function with dispatch. That's normally only used if (1) you need to dispatch multiple actions, or (2) you need access to the `getState` argument.

Here it looks like you could just do

    export function rejectEggFromBasket(args) {
       return {
         type: actionTypes.EGG_REJECTED,
         eggIndex: args.eggIndex,
         basketIndex: args.basketIndex
       };
    }

and ditto for `_acceptEggInBasket`. (It looks like maybe this was just left over from when you needed access to `getState` earlier.)

kswenson: You're right that I could have eliminated the thunk when I switched from using state to using args. In this case, I decided to leave it because I have a suspicion that when I implement the upcoming coin story that those two action creators will be dispatching the `CHALLENGE_COMPLETE` action as well.

PR #40, opened 2016-10-24
More robust animation implementation

kswenson:
Support dynamic computation of DOM target locations [#132334089]
Eliminate support for hatching drake in egg in clutch
Use higher-order components to simplify animation code
Move animatedComponents array into component state

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "react-router": "^2.8.1",
     "react-router-redux": "^4.0.5",
     "react-simpletabs": "^0.7.0",
+    "recompose": "^0.20.2",
     "redux": "^3.6.0",
     "redux-sounds": "^1.1.1",
     "redux-thunk": "^2.1.0",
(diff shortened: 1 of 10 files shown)


sfentress:
Looks very nice, and seems to clear up a bunch of code, which is a :+1:

Does it actually solve the resize/scroll issue, though? If you select an egg, and then either resize or (more common with small screens, perhaps) scroll, then select the basket, the animation goes to (and comes back to) the wrong place. Likewise if you scroll to see the popup message.

kswenson:
@sfentress Seems to be working for me. I had been testing mainly with resizing but it seems to be working with scrolling as well. Any chance you could be seeing a cached version?

sfentress:
Looks very nice, and also seems to clear up a bunch of code, which is a 👍 

(I notice that the animation still moves out of place if the user scrolls while the animation is ongoing. Could the handler that adds the `resize` listener also just add a `scroll` listener? Testing this, it seems to work out of the box with the two-line addition.)

kswenson:
Good point. I'll play around with adding scroll support. Thx

kswenson:
@sam I added the two lines you suggested for scroll support.

PR #41, opened 2016-10-25
Egg Game III: Authoring improvements and Case 4

kswenson:
Drakes/eggs specified explicitly in authoring and presented randomly
Add specification of Case 4:
- trial 1 (horns)
- trial 2 (armor)
- trial 3 (horns/wings/armor)


sfentress:
Looks good. :+1:. It would definitely be good to get some tests in, though...

PR #43, opened 2016-10-31
Egg game authoring improvements

kswenson:
- Author configuration should be JSON not JavaScript. [#133187035]
- Support author=upload URL parameter which allows user to load authoring document for testing [#133187467]
- Egg games use new-style dash allele strings for authoring [#133194823]
- Validate user-specified route changes
- Egg Game II: Don't generate the same target drake twice [Trudi]

@sfentress You're the logical reviewer for this one. I'll give the next one to Chris.

### DIFF
--- a/src/code/reducers/index.js
+++ b/src/code/reducers/index.js
@@ -13,20 +13,22 @@ import gametes from './gametes';
 import drakes from './drakes';
 import baskets from './baskets';
 
-const initialState = Immutable({
-  template: null,
-  hiddenAlleles: ['t','tk','h','c','a','b','d','bog','rh'],
-  trial: 0,
-  case: 0,
-  challenge: 0,
-  challenges: 1,
-  challengeProgress: {},
-  authoring: window.GV2Authoring,
-  endCaseUrl: urlParams.start
-});
+function initialState() {
+  return Immutable({
+            template: null,
+            hiddenAlleles: ['t','tk','h','c','a','b','d','bog','rh'],
+            trial: 0,
+            case: 0,
+            challenge: 0,
+            challenges: 1,
+            challengeProgress: {},
+            authoring: window.GV2Authoring,
+            endCaseUrl: urlParams.start
+          });
+}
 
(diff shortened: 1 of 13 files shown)


src/code/reducers/index.js
```diff
@@ -13,20 +13,22 @@ import gametes from './gametes';
...
-  hiddenAlleles: ['t','tk','h','c','a','b','d','bog','rh'],
-  trial: 0,
-  case: 0,
-  challenge: 0,
-  challenges: 1,
-  challengeProgress: {},
-  authoring: window.GV2Authoring,
-  endCaseUrl: urlParams.start
-});
+function initialState() {
```
kswenson: After converting `gv2.js` to `gv2.json` I ran into order-of-initialization issues between `initialState` and `window.GV2Authoring`. Rather than figuring out how to control the initialization order it seemed simplest to remove `initialState` from static initialization time.

sfentress: :+1:

kswenson:
Clearly it makes sense to convert all challenges to use the new dash allele format. My thinking at the time was that:

1. templates should be involved in the conversion because what fields need to be converted can be template-specific; and
2. if I converted the egg games (which I worked on) to the new format and then ran out of time before converting the other challenges, it wouldn't be too hard for someone else to convert the genome challenges, so a PT story could be written to do.

Your comments spurred me to take another look at this and the advantages of centralizing the conversion now seem to outweigh the value of giving templates control of the conversion process, so I've implemented a centralized conversion system and removed the other scattered conversions. If we wanted to give some measure of control back to the templates, we could add a static getPropertiesToConvert() function, but for now I'm just specifying the properties of interest. I also made the suggested changes to have the conversion function ignore inputs that don't look like dash allele strings.

sfentress:
This looks great. The only oversight, if I understand it correctly, is that an uploaded authored file won't get converted. Is that right, as it stands now?

Other than that, this all looks good to me.

PR #44, opened 2016-11-03
Egg Game unit test improvements

kswenson:
Add unit tests for
- new GeneticsUtils functions
  - alleleStringContainsAlleles()
  - isValidAlleleString()
  - isCompleteAlleleString()
- TrialGenerator
  - all-combinations
  - randomize-order
- Redux actions
  - changeBasketSelection()
  - changeDrakeSelection()
  - acceptEggInBasket()
  - rejectEggFromBasket()
  - navigateToCurrentRoute()
  - submitEggForBasket()

Sam got the last one, so this one's for @dstrawberrygirl.

### DIFF
--- a/test/actions/egg-rejected.js
+++ b/test/actions/egg-rejected.js
@@ -1,0 +1,70 @@
+import expect from 'expect';
+import * as actions from '../../src/code/actions';
+import reducer from '../../src/code/reducers/';
+
+const types = actions.actionTypes;
+
+describe('rejectEggFromBasket action', () => {
+  it('should create an action to reject the egg from the basket', () => {
+    const eggDrakeIndex = 0, basketIndex = 0, isChallengeComplete = false,
+          rejectEggArgs = { eggDrakeIndex, basketIndex, isChallengeComplete },
+          rejectEggAction = { type: types.EGG_REJECTED, eggDrakeIndex, basketIndex };
+
+    const dispatch = expect.createSpy();
+
+    actions.rejectEggFromBasket(rejectEggArgs)(dispatch);
+
+    expect(dispatch).toHaveBeenCalledWith(rejectEggAction);
+    expect(dispatch.calls.length).toBe(1);
+  });
+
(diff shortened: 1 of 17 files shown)


test/actions/egg-rejected.js
```diff
@@ -0,0 +1,70 @@
...
+    expect(dispatch.calls.length).toBe(1);
+  });
+
+  it('should create an action to reject the egg from the basket and show completion dialog', () => {
+    const eggDrakeIndex = 0, basketIndex = 0, isChallengeComplete = true,
+          rejectEggArgs = { eggDrakeIndex, basketIndex, isChallengeComplete },
+          rejectEggAction = { type: types.EGG_REJECTED, eggDrakeIndex, basketIndex },
+          showCompleteChallengeAction = {
+                                          type: types.MODAL_DIALOG_SHOWN,
+                                          message: "~ALERT.TITLE.GOOD_WORK",
```
dstrawberrygirl: Is this the correct message to show for a rejection?

kswenson: I agree that it seems a little weird, but Egg Game III is different than other challenges in that you're done when you finish the last egg, independently of whether you get it correct or not. First there's a SUBMIT_EGG action which triggers a correct/incorrect message. Then when you dismiss the initial feedback dialog, an EGG_ACCEPTED/EGG_REJECTED action is dispatched, either of which can result in finishing the challenge. So the "Good work!" message is for completing the challenge, not necessarily for sorting the last egg properly (for which feedback has already been given).

test/utilities/genetics-utils.js
```diff
@@ -69,6 +69,67 @@ describe("GeneticsUtils.convertDashAllelesObjectToABAlleles()", function() {
   });
 });
 
+describe("GeneticsUtils.alleleStringContainsAlleles()", function() {
+
```
dstrawberrygirl: I'm not seeing a case for if a null is passed in here - the other tests here do cover that.

kswenson: I'll add it.

PR #45, opened 2017-01-04
Update dependencies; add yarn support

kswenson:
- update dependencies (e.g. React 15.4.1)
- add yarn.lock file for yarn support


sfentress:
If we want to migrate to using yarn, should the readme be updated?

kswenson:
It's not a bad idea to mention yarn in the readme, but it isn't necessarily a migration. Yarn is designed to work with/alongside npm. Adding a `yarn.lock` file just means that if you type `yarn` you're guaranteed to get the exact set of libraries specified, whereas if you type `npm install` you'll get npm's usually-good-enough-but-not-completely-deterministic behavior. For day-to-day development either will generally be fine, i.e. personal preference can be used. Having the lock file means that deployments can be more predictable, however, so that's where using yarn could be recommended. I'll take a look at updating the readme and add a commit to this PR.

sfentress:
:+1:

PR #46, opened 2017-01-10
Fix GUIDE socket.io connection; show ITS messages

sfentress:
This fixes our communication to the ITS GUIDE Server, listens for ITS messages in their new format, and appends new messages to a new lightweight notification component.

Because the ITS messages are mostly demo strings for now, we don't show the messages by default. We can view the messages by including `?showITS=true` in the url. Eventually, we'll want to switch it so that the default is to view them.

https://www.pivotaltracker.com/story/show/137109565
https://www.pivotaltracker.com/story/show/137371291

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "axios": "^0.15.3",
     "babel-polyfill": "^6.20.0",
     "lodash": "^4.17.3",
+    "mustache": "^2.3.0",
     "react": "^15.4.1",
     "react-dom": "^15.4.1",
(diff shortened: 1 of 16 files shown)


sfentress:
@kswenson When you have a moment, can you please take a look?

Also I'm fixing the failing test and adding tests now.

sfentress:
This is now done. (Banged my head on the wall for a while trying to work out why `before` didn't seem to be defined, before finally realizing it was an eslint issue, not a mocha issue.)

PR #47, opened 2017-01-19
Formation of Gametes Game - Initial randomization/animation sequence [#136526365]

kswenson:
Implements the initial randomization/animation sequence which shows the formation of gametes from chromosomes:

For each chromosome pair of each parent:
- the selected chromosome initially alternates rapidly
- the pace of alternation slows before settling on a single chromosome (intended to evoke metaphor of slot machine animation)
- when a single chromosome is selected it animates to the half-genome in the center, as is true in other games when a chromosome is selected by the user

For each completed gamete of each parent:
- the chromosomes animate to the interior of the gamete image in the upper part of the screen
- then the entire gamete animates to the gamete pool at the bottom of the screen

The randomization/animation proceeds through the following stages, each of which produces one gamete for each parent:
- stage 1: Each pair of chromosomes for each parent randomizes individually before moving to the center upon selection. When the last chromosome is selected, the chromosomes animate to the gamete and then to the gamete pool.
- stage 2: All three chromosomes randomize simultaneously for one parent, then animate to the gamete and then to the gamete pool, after which the sequence is repeated for the other parent.
- stages 3-8: All three chromosomes randomize simultaneously for both parents, then animate to their respective gametes simultaneously, then animate to their respective gamete pools simultaneously.

At the last demo/planning meeting it was decided that once the slot machine animation was in place (and the fading opacities changed) it should be made available for feedback and code review. The current state of the Formation of Gametes Game can be tested at http://concord-consortium.github.io/geniblocks/gv2/#/2/2. Ignore the layout issues in the gamete pools at the bottom of the page; I'll be working on those next.

Under the hood the slot machine animation timing is handled with a pair of utility classes (RepeatTimer, TimerSet) which manage the asynchronous timing of the randomization. The animations are handled by the animationEvents model that was in place previously with some minor changes (e.g. each event has its own completion routine rather than a single shared completion routine). I did some research into other animation libraries and think that a React version of something like GreenSock's timeline capabilities would be a nice improvement, but it doesn't seem to exist and I can't quite see my way clear to implementing it (yet).

### DIFF
--- a/src/code/templates/egg-game.js
+++ b/src/code/templates/egg-game.js
@@ -600,6 +600,7 @@ export default class EggGame extends Component {
   }
 
   componentWillMount() {
+    _this = this;
     challengeDidChange = true;
     resetAnimationEvents(false, this.props.showUserDrake);
     animatedComponents = [];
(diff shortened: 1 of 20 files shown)


src/code/templates/egg-game.js
```diff
@@ -101,44 +185,375 @@ var animationEvents = {
...
+
+      function toggleAnimatingGameteChromosome(sex, chromName, sides) {
+        let animatingGametes = _this.state.animatingGametes || [{}, {}, {}, {}],
+            gamete = animatingGametes && animatingGametes[sex],
+            currSide = gamete && gamete[chromName],
+            newSide = currSide != null
+                        ? (currSide === sides[0] ? sides[1] : sides[0])
+                        : oneOf(sides);
+        animatingGametes[sex] = assign({}, gamete, { [chromName]: newSide });
+        if (_this) _this.setState({ animatingGametes });
```
sfentress: I think that the `if (_this)` pattern is pretty close to the "isMounted" antipatern described here: https://facebook.github.io/react/blog/2015/12/16/ismounted-antipattern.html, where they recommend explicitly unsubscribing from any timers or listeners in `componentWillUnmount` instead.

Testing aborted animations (by changing the challenge hash in the url bar in the middle of animations), there are still a number of errors that get thrown, even with this check. For example, in this function, the `_this.state` above is null. Also, there are many cases where `setTimeout(() => [...] _this.setState` is used without checking `_this`.

I was worried that unsubscribing from all the animation events would be a really ugly task, and that this pattern would be better on balance, but actually your `timerSet` makes this really easy. As far as I could tell, I could remove all the instances of `if (_this)` and remove all the unmounting errors with just a few changes (which I could also put in a commit, here or in another branch, if that's easier):

1. Move `timerSet` to the top-level scope, and call `if (timerSet) timerSet.reset();` in `resetAnimationEvents`
2. Add a quick utility (at the top scope, or in /utilities) for remembering and clearing all timeouts:
```
var timers = [];
function _setTimeout(fn, delay) {
  timers.push(setTimeout(fn, delay));
}
function clearTimeouts() {
  for (let timer of timers) {
    clearTimeout(timer);
  }
}
```
3. Replace all `setTimeout` with `_setTimeout`, and call `clearTimeouts();` in `resetAnimationEvents`.

It's quite possible I missed something, but in testing I found that with those changes I could exit in the middle of any animation and not get errors.

Thoughts?

src/code/templates/egg-game.js
```diff
@@ -13,6 +13,12 @@ import ChromosomeImageView from '../components/chromosome-image';
 import TimerSet from '../utilities/timer-set';
 import t from '../utilities/translate';
 
+// debugging options to shorten animation/randomization for debugging purposes
+// IMPORTANT: remember to set these back to their defaults (e.g. false, false, 0) before committing
+const debugSkipIntroGameteAnimation = false,  // set to true to skip intro gamete animation
+      debugSkipRandomGameteAnimation = false, // set to true to skip randomized gamete animation
+      debugTotalGameteCount = 0;              // non-zero value stops randomization after # gametes
```
sfentress: Nice :)

I can imagine that eventually we'll want a bunch of debugging cheat codes, which perhaps look up values on `window` so that we can set them directly from the command line. (If we ever do, though, we'll need to obfuscate them to prevent students from discovering them...)

src/code/templates/egg-game.js
kswenson: I think your approach is a good one. I was/am familiar with the discussion around isMounted being an anti-pattern, but also think that discussion glosses over the fact that many asynchronous actions can't easily be interrupted. In our case, however, we certainly have the ability to interrupt our timer tasks and I would expect react-motion to be handling the lifecycle events appropriately for its own animations. (I think I was unconvinced of that at the time I was adding the `if (_this)` tests.) If you have some working code then adding a commit to the PR would be great. Thanks

sfentress: Ok, take a look at my commit and see what you think.

For me, this is :+1: for merging. I don't know whether you want to put your gamete-selection features in a new PR, or continue this one, either is fine with me.

kswenson:
Code in your commit looks good. In reviewing it I noticed a few additional magic animation constants in my code that I should move to the top of the file. I'll do a bit of testing and assuming everything holds up as expected we can call it wrap. Thanks

PR #48, opened 2017-01-25
Authored linked genes

sfentress:
This adds the ability to include

```
    "linkedGenes": {
        "drakes": [0, 1],
        "genes": "tail, horns, color, black, dilute, armor, nose, bogbreath"
      }
```

in the authoring, to define genes may not be explicitly authored, but must remain identical between two (or more) drakes.

https://www.pivotaltracker.com/story/show/136137281


sfentress:
@kswenson If you can look at this when you can, that would be great.

As far as I can tell, the Travis failure is not a real failure (i.e. is not related to this PR).

PR #51, opened 2017-01-26
User can select gametes in pool; fertilize button enabled when two gametes selected

kswenson:
- Users can select a gamete by clicking on it in a pool [#136526521]
- Selected gametes highlight their corresponding parent chromosomes and show their chromosomes in the center [#136526521]
- User selection of gamete (or clicking in pool) ends animation and completes pool
- Support showing scaled chromosomes in gametes in pool always, never, or when selected (default)
- With two gametes selected, user can fertilize to create an offspring drake [#136526581]
- If offspring drake is kept, selected gametes are removed from the pool

Under the hood there's a bit of refactoring of the redux code. The `actionTypes` and ITS constants are split out into separate files, and the new code for dealing with the `parentGametes` part of the state is encapsulated in a single module following the "ducks" model rather than being separated into actions.js, etc.


src/code/modules/parent-gametes.js
```diff
@@ -0,0 +1,134 @@
...
+
+/*
+ * selectors
+ */
+export const fatherGametePool = state => state.pools[0];
+export const motherGametePool = state => state.pools[1];
+export const gametePoolSelector = sex => sex ? motherGametePool : fatherGametePool;
+export const fatherSelectedGameteIndex = state => state.selectedIndices[0];
+export const motherSelectedGameteIndex = state => state.selectedIndices[1];
+export const fatherSelectedGamete = createSelector(
```
sfentress: I can't find where `fatherSelectedGamete` and `motherSelectedGamete` are used.

src/code/action-types.js
```diff
@@ -0,0 +1,33 @@
+const actionTypes = {
+  SESSION_STARTED: "Session started",
+  AUTHORING_CHANGED: "Authoring changed",
+  LOADED_CHALLENGE_FROM_AUTHORING: "Loaded challenge from authoring",
+  NAVIGATED: "Navigated",
+  CHALLENGE_COMPLETE: "Challenge completed",
+  BRED: "Bred",
+  ALLELE_CHANGED: "Allele changed",
+  SEX_CHANGED: "Sex changed",
+  GAMETE_CHROMOSOME_ADDED: "Gamete chromosome added",
```
sfentress: It seems like this could be part of `parent-gametes` too, right?

src/code/reducers/index.js
```diff
@@ -10,7 +10,7 @@ import moves from './moves';
 import modalDialog from './modal-dialog';
 import userDrakeHidden from './user-drake-hidden';
 import gametes from './gametes';
-import gametePools from './gamete-pools';
```
sfentress: should the `gametes-pools` reducer be deleted?

src/code/modules/parent-gametes.js
```diff
@@ -0,0 +1,134 @@
...
+/*
+ * action creators
+ */
+export function addGametesToPool(sex, gametes) {
+  return {
+    type: GAMETES_ADDED_TO_POOL,
+    sex,
+    gametes: cloneDeep(gametes),
+    meta: {
+      logTemplateState: true,
```
sfentress: I realize this property isn't well documented (I had to do a quick search through the code to remind myself what it does), but it relies on the template having a `logState` function, which returns template-specific, ITS-meaningful info.

You can see the one example in `genomeChallenge`.

Without the template implementing `logState`, this variable being `true` doesn't actually do anything.

As an explanation, the ITS folks want state data, but they don't want our entire Redux state object, so instead we translate our state into ITS-meaningful state.

sfentress:
I think this is fine. The one bug that I've noticed is that, if the user selects any parent chromosome by themselves while the animation is going on, the chromosome will be selected, and the subsequent animation will no longer play, and the challenge gets into a weird state.

This would presumably be an easy fix by making the reducer for `GAMETE_CHROMOSOME_ADDED` do nothing in this interactionType (and that action and reducer could be moved to the module, I'd think).

Besides that, there are some non-urgent comments, but this can all be merged in when you think it's ready.

src/code/modules/parent-gametes.js
kswenson: I was just following the pattern set out by whatever other action I copied initially. I guess I'll set it to false for now and we can figure out what logState should do at a later date.

src/code/action-types.js
kswenson: That action existed before the parentGametes module and is mediated by the separate gametes reducer. I thought about combining the gametes reducer/actions into the parentGametes module but doing so didn't seem like it was necessarily the highest priority.

src/code/modules/parent-gametes.js
kswenson: They may not be used at the moment. I was thinking about what the module API should look like and it seemed reasonable that clients might want to have encapsulated access to the gametes themselves. (Plus, it gave me a chance to see how selector composition worked with the reselect library, which I thought would be a good thing to understand. 😉 )

src/code/reducers/index.js
kswenson: Good catch

src/code/action-types.js
kswenson: Done

PR #52, opened 2017-01-27
Randomize trials

sfentress:
This adds a generic way to randomize trials whenever an author is creating a list of predefined trials (e.g. the targetDrakes list for the genomeChallenge). It works by creating a shuffled array of trial-numbers, and passing that value in, instead of the actual trial number, whenever we use the trial number to calculate state (currently just in `authoredDrakesToDrakeArray`).

This also adds the ability to define an array of initial drakes for the genome challenge.

### DIFF
--- a/src/code/reducers/helpers/load-state-from-authoring.js
+++ b/src/code/reducers/helpers/load-state-from-authoring.js
@@ -50,6 +50,31 @@ function processAuthoredDrakes(authoredChallenge, trial, template) {
   return drakes;
 }
 
+// Returns an array [0...len], optionally shuffled
+function createTrialOrder(trial, trials, currentTrialOrder, shuffle) {
+  if (trial > 0) {
+    return currentTrialOrder;
+  }
+  if (!trials || trials.length === 0) {
+    return [0];
+  }
+
+  let trialOrder = [];
+  for (let i = 0, ii = trials.length; i < ii; i++) {
+    trialOrder[i] = i;
+  }
+  if (shuffle) {
+    var j, x, i;
+    for (i = trialOrder.length; i; i--) {
+        j = Math.floor(Math.random() * i);
+        x = trialOrder[i - 1];
+        trialOrder[i - 1] = trialOrder[j];
+        trialOrder[j] = x;
+    }
+  }
+  return trialOrder;
+}
+
 export function loadStateFromAuthoring(state, authoring, progress={}) {
   let trial = state.trial ? state.trial : 0;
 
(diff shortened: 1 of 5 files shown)


src/code/templates/genome-challenge.js
```diff
@@ -84,6 +84,8 @@ export default class GenomeChallengeTemplate extends Component {
   static authoredDrakesToDrakeArray = function(authoredChallenge, trial) {
     if (authoredChallenge.trialGenerator) {
       return generateTrialDrakes(authoredChallenge.trialGenerator, trial);
+    } else if (authoredChallenge.initialDrake.constructor === Array) {
```
kswenson: Array.isArray() is more robust and is supported by all browsers we care about (cf. [Determining with absolute accuracy whether or not a JavaScript object is an array](http://web.mit.edu/jwalden/www/isArray.html)).

src/code/reducers/helpers/load-state-from-authoring.js
```diff
@@ -50,6 +50,31 @@ function processAuthoredDrakes(authoredChallenge, trial, template) {
...
+    return currentTrialOrder;
+  }
+  if (!trials || trials.length === 0) {
+    return [0];
+  }
+
+  let trialOrder = [];
+  for (let i = 0, ii = trials.length; i < ii; i++) {
+    trialOrder[i] = i;
+  }
```
kswenson: Could also use Lodash's `times` function, something like:
```
let trialOrder = times(trials.length, index => index);
```

src/code/reducers/helpers/load-state-from-authoring.js
```diff
@@ -113,34 +140,9 @@ export function loadNextTrial(state, authoring, progress) {
...
-    hiddenAlleles,
-    baskets,
-    drakes,
-    trial,
-    correct: 0,
-    errors: 0,
-    moves: 0,
-    goalMoves,
-    userDrakeHidden: true,
-    challengeProgress: progress
```
kswenson: I _love_ how much code you managed to eliminate here as part of this cleanup. 😉

src/code/reducers/helpers/load-state-from-authoring.js
```diff
@@ -113,34 +140,9 @@ export function loadNextTrial(state, authoring, progress) {
   if (state.trialSuccess){
     trial = (state.trial < state.trials.length) ? (trial+1) : 1;
```
kswenson: You didn't change this line, so this comment is not technically related to this PR, but if the point of this line is to wrap around, shouldn't it wrap to trial 0?

sfentress: Agree. I hadn't seen this line before.

PR #56, opened 2017-02-02
Fertilize the Egg game

kswenson:
@sfentress Final commit added which wraps up all existing stories (to this point). You should be able to just review the most recent commits since you last reviewed the code.

Update: most recent commits wrap up basic functionality that was missing at the point the PR was first submitted:
- gametes are generated at the beginning of each trial/challenge and animation just syncs to pre-generated gametes
- generated gametes are guaranteed to include ones that can match the target
- gamete counts can be specified on a per-trial basis
- gamete counts are specified as integers rather than sets of `{}`s
- if there's only one gamete for a given parent it is selected automatically

@sfentress You may review the changes at your convenience.

Add challenges to authoring document
- Dragon 3.1: Draco (piece 2): #5/2 - Egg Game I(g) (select gametes to create five unique drakes)
- Dragon 3.1: Draco (piece 3): #5/3 - Egg Game II(g) (select gametes for breeding to match drakes)
- Currently, all trials act like Trial 3 in the spec document (4 gametes for mother, 4 gametes for father) [#138978405]
- Layout tweaks for Fertilize the Egg game
- Number of gametes is authorable and mother/father can have different number of gametes
- Gametes animate to correct location in pool with arbitrary numbers of gametes
- Fertilize the Egg game supports multiple trials

Note: The random selection of the target is currently unrelated to the random selection of the gametes, i.e. there's no guarantee that there is a combination of gametes that will generate a given target drake. That will obviously be fixed in an upcoming Pull Request.

Update examples page to match Trudi's new order

Bug-fixes
- Add default value for options argument of resetAnimationEvents()
- Fix PropTypes declaration for `showChromosomes`

### DIFF
--- a/src/resources/authoring/gv2.json
+++ b/src/resources/authoring/gv2.json
@@ -300,7 +300,7 @@
         "alleles": "w-W, m-m, T-T, h-h, A1-A1, C-C, B-B, D-D, rh-rh, Bog-Bog",
         "sex": 0
       },
-      "gametes": [[{}, {}, {}, {}, {}, {}, {}, {}], [{}, {}, {}, {}, {}, {}, {}, {}]],
+      "gameteCounts": [8, 8],
       "visibleGenes": "metallic, wings, forelimbs, hindlimbs"
     },
     {
@@ -318,7 +318,7 @@
         "alleles": "w-W, m-m, T-T, h-h, A1-A1, C-C, B-B, D-D, rh-rh, Bog-Bog",
         "sex": 0
       },
-      "gametes": [[{}, {}, {}, {}], [{}, {}, {}, {}]],
+      "gameteCounts": [[2, 1], [1, 3], [4, 4]],
       "targetDrakes": [{}, {}, {}],
       "visibleGenes": "metallic, wings, forelimbs, hindlimbs"
     }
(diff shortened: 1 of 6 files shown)


src/resources/authoring/gv2.json
```diff
@@ -283,6 +283,7 @@
         "alleles": "w-W, m-m, T-T, h-h, A1-A1, C-C, B-B, D-D, rh-rh, Bog-Bog",
         "sex": 0
       },
+      "gametes": [[{}, {}, {}, {}, {}, {}, {}, {}], [{}, {}, {}, {}, {}, {}, {}, {}]],
```
sfentress: This pattern started as a bit of a hack from the pre-existing arrays of authored target drakes, saying "There could be some target drakes here, but we don't need to author them."

I feel like if there are 16 `{}`s in a row, we could say that the hack has outlived its usefulness. :)

Since the authoring json doesn't need to directly mirror the state json, should this just change to two `int`s?

kswenson: I did give it some thought and probably would have changed it if not for the feeling that Trudi's going to want to have more control over the set of gametes generated, and so we're likely to end up with a similar-looking array of more fully-specified gametes whose order gets randomized at run-time. But there's no guarantee of that, and I agree that long strings of `{}`s in a row should not be the API we end up with.

sfentress: Sounds good to me. Feel free to merge. :+1:

kswenson:
Eli and I reviewed the remainder of the PR in a side-by-side review.

PR #57, opened 2017-02-09
Tutorial pen and consistent sex-linked traits

ekosmin:
@kswenson @sfentress These commits first change the "sandbox" tutorial level into a "create 5 drakes" level which includes a drake pen. This change uses many of the same actions as the other template which includes a pen- the main difference is that here the baby drake is not removed when it is added to the pen.

The other fix is to address issues with sex-linked traits changing as a dragon's sex is changed. Now, as in GV1.0, when a drake changes from female to male, it records any alleles it is losing in the Redux state. When the drake changes from male to female, it restores any alleles so that it returns to the same state as the original female drake.

One difference from GV1.0 is that you can now fully specify an initial drake, regardless of its sex. By over-specifying a male (i.e. providing a full set of sex linked alleles), you specify how the drake will look if converted to a female. This happens during the initialization of the drakes array, by comparing the male drake to a female drake created with the same authored alleles.

### DIFF
--- a/src/code/actions.js
+++ b/src/code/actions.js
@@ -554,7 +554,7 @@ export function hatch() {
 }
 
 
-function _keepOffspring(index, success, interactionType, keepDisplayBaby) {
+function _keepOffspring(index, success, interactionType, shouldKeepSourceDrake) {
   let incrementMoves = !success;
   return {
     type: actionTypes.OFFSPRING_KEPT,
@@ -562,15 +562,15 @@ function _keepOffspring(index, success, interactionType, keepDisplayBaby) {
     index,
     success,
     incrementMoves,
-    keepDisplayBaby
+    shouldKeepSourceDrake
   };
 }
 
-export function keepOffspring(index, success, maxDrakes, keepDisplayBaby) {
+export function keepOffspring(index, success, maxDrakes, shouldKeepSourceDrake) {
   return (dispatch, getState) => {
     const { interactionType } = getState();
 
-    dispatch(_keepOffspring(index, success, interactionType, keepDisplayBaby));
+    dispatch(_keepOffspring(index, success, interactionType, shouldKeepSourceDrake));
 
     if (success) {
       const { drakes } = getState();
@@ -583,7 +583,7 @@ export function keepOffspring(index, success, maxDrakes, keepDisplayBaby) {
         explanation: "~ALERT.DUPLICATE_DRAKE",
         rightButton: {
           label: "~BUTTON.TRY_AGAIN",
-          action: keepDisplayBaby ? "dismissModalDialog" : "resetGametes"
+          action: shouldKeepSourceDrake ? "dismissModalDialog" : "resetGametes"
         }
       }));
     }
--- a/src/code/actions.js
+++ b/src/code/actions.js
@@ -554,22 +554,23 @@ export function hatch() {
 }
 
 
-function _keepOffspring(index, success, interactionType) {
+function _keepOffspring(index, success, interactionType, keepDisplayBaby) {
   let incrementMoves = !success;
   return {
     type: actionTypes.OFFSPRING_KEPT,
     interactionType,
     index,
     success,
-    incrementMoves
+    incrementMoves,
+    keepDisplayBaby
   };
 }
 
-export function keepOffspring(index, success, maxDrakes) {
+export function keepOffspring(index, success, maxDrakes, keepDisplayBaby) {
   return (dispatch, getState) => {
     const { interactionType } = getState();
 
-    dispatch(_keepOffspring(index, success, interactionType));
+    dispatch(_keepOffspring(index, success, interactionType, keepDisplayBaby));
 
     if (success) {
       const { drakes } = getState();
@@ -582,7 +583,7 @@ export function keepOffspring(index, success, maxDrakes) {
         explanation: "~ALERT.DUPLICATE_DRAKE",
         rightButton: {
           label: "~BUTTON.TRY_AGAIN",
-          action: "resetGametes"
+          action: keepDisplayBaby ? "dismissModalDialog" : "resetGametes"
         }
       }));
     }
--- a/src/code/actions.js
+++ b/src/code/actions.js
@@ -566,15 +566,28 @@ function _keepOffspring(index, success, interactionType, shouldKeepSourceDrake) 
   };
 }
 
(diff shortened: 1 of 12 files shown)


ekosmin:
I realize I had to update the expected state in a few tests to get them to pass. One issue I ran into was that the initial "secondXAlleles" info for male drakes is random without specification, which made testing hard. I hacked around this by making all the drakes female- I'll need some help to fix the tests back up

test/templates/egg-game.js
```diff
@@ -106,28 +106,33 @@ describe('loading authored state into template', () => {
           "drakes": [
             {
               "alleleString": nextState.drakes[0].alleleString, // we check valid alleles below
+              "secondXAlleles": null,
```
sfentress: In theory this could be skipped in all these tests -- and the state cleaned up a bit -- if `secondXAlleles` were only set if it were non-null in `load-state-from-authoring`. However, I think we have other `null` values scattered about the state, and we don't have a policy on removing them, so I think this is fine.

src/code/templates/genome-playground.js
```diff
@@ -20,8 +26,19 @@ export default class GenomeContainer extends Component {
       onSexChange(0, newSex);
     };
 
-    const handleAdvanceChallenge = function() {
-      onCompleteChallenge();
+    const handleSubmit = function() {
+      let child = new BioLogica.Organism(BioLogica.Species.Drake, drakes[0].alleleString, drakes[0].sex),
+          childImage = child.getImageName(),
```
sfentress: Instead of creating a new child here, since it's just a clone of the main drake, couldn't you just do `let childImage = drake.getImageName()`?

(You'd have to rename the clashing `let drake` below, though.)

freichsman:
Question: if the over-specified male drake is heterozygous, is the first
allele of the two being specified the one that the male will get?

Question: is this email thread the right place to ask a question like this?

ekosmin:
Great questions! That's exactly right, the first of the alleles specified
will be used by the male, I just confirmed this behavior. So, an rh-Rh male
will not have a nose spike while a Rh-rh male will.

I would probably opt for asking these questions in a more "public" place
like PT so the answers are recorded for the future and could be viewed by
others. This is the first open-source project I've worked on though, so I'm
definitely not sure about best practices.

Eli

ekosmin:
Never mind, I didn't realize email comments here are also posted publicly
to GitHub! So this is probably also a fine place to ask them.

ekosmin:
Thanks to @kswenson and @sfentress again for your feedback! These commits should take care of all the comments. There are two main pieces I'd like a second pair of eyes on:

I ended up moving the code that adds unique drakes to the pen into the `keepOffspring` action and out of `egg-game` and `genome-playground`- the action now checks for success on its own. To do this, I pass in a list of indices where the `keptDrakes` can be found in the `drakes` array. I figured it would be cleaner to pass around the indices than the actual drakes...let me know how that code looks, especially if there's any ES6 ways I could spruce it up.

To Sam's good point, I also snuck in a unit test to check for proper initialization of secondXAlleles for male drakes into the `genome-challenge` tester. It creates a drake with specified sex-linked chromosomes, and makes sure they're loaded correctly into the state. As discussed with Kirk, these tests really belong in a separate file for `load-state-from-authoring` tests, but I don't want to hold up the merge for that.

src/code/actions.js
```diff
@@ -566,15 +566,28 @@ function _keepOffspring(index, success, interactionType, shouldKeepSourceDrake)
...
-export function keepOffspring(index, success, maxDrakes, shouldKeepSourceDrake) {
+export function keepOffspring(index, keptDrakesIndices, maxDrakes, shouldKeepSourceDrake) {
   return (dispatch, getState) => {
-    const { interactionType } = getState();
+    const { interactionType, drakes } = getState();
+
+    let offspringOrg = new BioLogica.Organism(BioLogica.Species.Drake, drakes[index].alleleString, drakes[index].sex),
+        success = true;
+    // Don't keep any drakes with the same phenotypes
+    keptDrakesIndices.forEach(function(keptDrakeIndex) {
```
sfentress: This is fine, since it's a small operation, but `every` might be a better function to use here, because it won't keep spinning around after you've already set `success = false`.

```
success = keptDrakesIndices.every(function(keptDrakeIndex) {
...
return keptImage !== offspringOrg.getImageName();
});
```

sfentress:
This looks great. I think the `keepOffspring` change is good.

Now that `keepOffspring` is more complex, it would be great to have a test for it.

As for whether we need to have a specific file for `load-state-from-authoring`, I'm not sure. `load-state-from-authoring` is pretty dependent on the specific template that the game is loading, particularly the drakes array. That's why we now have a few template tests -- it's mostly to test `load-state-from-authoring`. It's true that some things are cross-cutting and happen independently of templates, like how drakes are first instantiated, but for the moment at least I think it's fine keeping them in the template tests, as you have done.

src/code/reducers/drakes.js
```diff
@@ -24,16 +24,15 @@ export default function drakes(state = initialState, action) {
           secondXAlleles = GeneticsUtils.computeExtraAlleles(oldOrg, newOrg);
           alleleString = newOrg.getAlleleString();
         }
-        else if (drakeDef.sex === BioLogica.MALE && action.newSex === BioLogica.FEMALE) {
+        else if (drakeDef.sex === BioLogica.MALE && action.newSex === BioLogica.FEMALE && drakeDef.secondXAlleles) {
```
kswenson: With this change I don't think the `alleleString` gets set when `drakeDef.secondXAlleles` is empty. The test for `drakeDef.secondXAlleles` should be moved inside the braces. You might consider taking a TDD approach: write a unit test for this case before changing the code.

src/code/actions.js
ekosmin: Nice, never knew about that method, thanks for the tip!

src/code/reducers/drakes.js
ekosmin: That's a nice catch Kirk- you're right that there should be a test for this if I'm going to say here that an empty `secondXAlleles` is valid. It definitely would have caught this issue.

I've never had the chance to try out TDD- it would be nice to do that for my next project.

ekosmin:
Thanks Kirk! I'll go ahead and merge this, but please do let me know if you or @sfentress have any thoughts on the style of my new tests for keepOffspring. I tried to pull out some constants into a convenient place, and also used a spy with a return value.

PR #59, opened 2017-02-14
Update allele labels

kswenson:
Fixes [#139578683]
- "colored" --> "color"
- "black"/"brown" --> "gray"/"orange"
- "metallic"/"nonmetallic" --> "shiny"/"dull"

### DIFF
--- a/src/code/app.js
+++ b/src/code/app.js
@@ -95,6 +95,16 @@ function handleCompleteUpload(authoring) {
   renderApp();
 }
 
+function initializeApp() {
+  // BioLogica overrides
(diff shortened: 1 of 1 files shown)


sfentress:
Frieda reports that time is of the essence, and that it's not worth distracting from the main concern of implementing the layouts, so I'm ok with this as a temporary solution :+1:

Feel free to merge.

kswenson:
Given @sfentress's point that we've already started down the path of updating the allele strings in `BioLogica.js` it doesn't make sense to have some of the updates in `BioLogica.js` and some of them as overrides in GV2, so I'm closing this in favor of a `BioLogica.js` pull request to follow shortly.

PR #60, opened 2017-02-21
Adding component for viewing parents in the egg game

ekosmin:
Here's the new drake view! I hard-coded a lot of pieces here, including text and orientation (based on organism sex). I also threw the new font definition into a common file. Any feedback is appreciated

screen shot 2017-02-21 at 3 21 13 pm


src/stylus/parent-drake.styl
```diff
@@ -0,0 +1,24 @@
...
+			margin-right: 20px
+
+	&.male
+		background: url('../resources/images/male_drake_overlay.png')
+		justify-content: flex-start
+		.organism
+			margin-left: 20px
+
+	.parentTitle
+		margin-top: 10px
```
kswenson: It appears that this file uses embedded tabs rather than spaces. Elsewhere we have used spaces rather than tab characters and so this convention should be followed with new files.

ekosmin: I've updated my Sublime preferences since then, but I guess this file slipped through. Thanks for the catch!

ekosmin:
Abandoning for pull request from fv-parent-view

PR #63, opened 2017-02-23
Fv breed button

ekosmin:
Creating the breed button, which is just a few nested divs with an onClick function passed in. Waiting on a separate image for the disabled view.


src/code/containers/challenge-container-selector.js
```diff
@@ -24,7 +24,7 @@ function mapContainerNameToContainer(containerName) {
 class ChallengeContainerSelector extends Component {
 
   static propTypes = {
-    authoring: PropTypes.object.isRequired,
+    authoring: PropTypes.array.isRequired,
```
ekosmin: This was throwing errors before I changed it- `authoring` *is* an array, so I can see why this is correct, but I don't know why it wasn't a problem when these propTypes were in `challenge-container`

kswenson: Good catch. I should have noticed that myself.

src/code/fv-components/breed-button.js
```diff
@@ -0,0 +1,31 @@
+import React, {PropTypes} from 'react';
+import classNames from 'classnames';
+import t from '../utilities/translate';
+
+const BreedButtonView = React.createClass({
```
kswenson: Seems like this could have been implemented as a stateless functional component.

src/code/fv-components/breed-button.js
```diff
@@ -0,0 +1,31 @@
...
+
+  handleClick: function() {
+    this.props.onClick();
+  },
+
+  render: function() {
+    return (
+      <div className='geniblocks breed-button'>
+        <div className={classNames('clickable-button', { 'disabled': this.props.disabled})} onClick={this.handleClick}>
+          <div className="text unselectable">{ t("~FV_EGG_GAME.BREED_BUTTON").toUpperCase() }</div>
```
kswenson: I would leave off the call to `toUpperCase` here. The author of the string should be able to choose whether the string is capitalized or not without having the code override the author's choice. I'll comment on class names in the CSS file where they're used.

src/code/utilities/lang/en-us.js
```diff
@@ -1,5 +1,8 @@
 export default {
   // Challenge instructions
+  "~FV_EGG_GAME.BREED_BUTTON": "Breed",
```
kswenson: Already mentioned above, but since the design has the label capitalized, I think it should be capitalized here rather than in code.

src/stylus/fablevision/breed-button.styl
```diff
@@ -0,0 +1,27 @@
+.geniblocks.breed-button
```
kswenson: For clarity, I would call this `breed-button-surround`, because it's not the actual button.

src/stylus/fablevision/breed-button.styl
```diff
@@ -0,0 +1,27 @@
...
+  background: url('../resources/fablevision/layout-c/breed_surround.png')
+  width: 370px
+  height: 369px
+  display: flex
+  justify-content: center
+  align-items: center
+
+  .clickable-button
```
kswenson: `clickable-button` implies that this is applicable to buttons in general, but since it's really breed button specific, I'd call it `breed-button` (once the current `breed-button` is renamed to `breed-button-surround`.

src/stylus/fablevision/breed-button.styl
```diff
@@ -0,0 +1,27 @@
...
+    align-items: center
+    background: url('../resources/fablevision/layout-c/breed_button_UNCLICKED.png')
+    cursor: pointer
+
+    &:active
+    &.disabled
+      cursor: default
+      background: url('../resources/fablevision/layout-c/breed_button_CLICKED.png')
+
+    .text
```
kswenson: `text` implies that it's applicable to all text. I would suggest something like `button-label` which specifies the `font-family`, since all buttons are likely to use the same font, and then another class for `breed-button-label` which specifies the `font-size`, since that's likely to be unique to the breed button.

src/code/containers/challenge-container-selector.js
kswenson: One additional thought on this topic: seems like an opportunity for an automated test. It would be more of an acceptance test than a unit test, but one could imagine instantiating all challenges with traps for propTypes warnings and other such failures. Details on how to go about it at [Make React PropType warnings throw errors with enzyme.js + sinon.js + mocha.js](https://gist.github.com/scmx/d98cc058a7c3dfef7890) and [Testing React PropTypes with sinon](http://stackoverflow.com/questions/40859389/testing-react-proptypes-with-sinon). Not something to hold up this PR, but worth thinking about moving forward.

src/stylus/fablevision/breed-button.styl
ekosmin: Thanks for all these comments Kirk! I wrote it like this originally because I was thinking about it just from the perspective of this file, where it's clear that `text` is part of a breed button. In my new commit, I took your suggestion and moved `button-label` to its own class, and defined it in `challenges.styl` so all labels have the correct `font-family`. I can see how this approach will reduce the amount of CSS I write going forward- let me know if I have the right idea.

PR #66, opened 2017-02-24
Updating URL scheme to have 3 parts for level, mission and challenge

ekosmin:
The URL scheme now matches the master development document. This required adding `level` to the URL scheme, moving around the authoring doc, and updating some tests.


src/code/actions.js
```diff
@@ -32,12 +32,13 @@ export function changeAuthoring(authoring) {
   };
 }
 
-export function navigateToChallenge(_case, challenge) {
+export function navigateToChallenge(level, _case, challenge) {
```
ekosmin: I left all `case` references alone for now, even though it should technically be `mission` now. Task for developer cleanup?

src/resources/authoring/gv2.json
```diff
@@ -1,397 +1,424 @@
...
-        {
-          "alleles": "A1-A1, H-H, Bog-, D-, rh-rh",
+    ]
+  ],
+  [
+    [
+      {
+        "comment": "*** LEVEL STAND IN ***",
+        "template": "GenomeChallenge",
+        "instructions": "THIS LEVEL NOT YET COMPLETE",
```
ekosmin: Currently, we don't have any challenges complete from level 3, so I added this fake challenge so that the numbering for level 4 would be correct. Ideally, I would update the number-wrapping logic to process empty challenges, but I figure these bogus challenges work alright for now.

kswenson: I think it would make sense to have a `StubTemplate` (or some other more suitable name) which could just put up a message that included the comment and a "Not Yet Implemented" message. With that in place, we wouldn't need to redo the number-wrapping logic.

src/code/actions.js
kswenson: One other thought: consider naming the outer object `routeSpec` which disentangles the name of the outer object from the terminology of its inner components.

src/code/components/challenge-award.js
```diff
@@ -10,7 +10,7 @@ class ChallengeAwardView extends React.Component {
   };
 
   static defaultProps = {
-     challengeAwards: {"caseId":0, "challengeId":0, "challengeCount":0, "progress":[]},
+     challengeAwards: {"missionId":0, "challengeId":0, "challengeCount":0, "progress":[]},
```
ekosmin: This is the only part I'm not sure about- I renamed `caseId` to `missionId`, but I didn't add `levelId` anywhere in this class. The awards seem to work just fine, even though I might expect there to be an issue with duplicate IDs without the level number.

test/actions/egg-accepted.js
```diff
@@ -31,15 +31,15 @@ describe('acceptEggInBasket action', () => {
...
+                                            label: "~BUTTON.NEXT_MISSION",
                                             action: "navigateToNextChallenge"
                                           },
                                           top: undefined,
                                           showAward: true
                                         };
 
     const dispatch = expect.createSpy();
-    const getState = () => ({case: 0, challenge: 0, trial: 0, trials: [{}], authoring: [[{}]] });
+    const getState = () => ({routeSpec: {mission: 0, challenge: 0}, trial: 0, trials: [{}], authoring: [[{}]] });
```
kswenson: Seems like `level` should be included in the `routeSpec`, even though it's apparently not strictly necessary.

test/actions/egg-rejected.js
```diff
@@ -31,15 +31,15 @@ describe('rejectEggFromBasket action', () => {
...
+                                            label: "~BUTTON.NEXT_MISSION",
                                             action: "navigateToNextChallenge"
                                           },
                                           top: undefined,
                                           showAward: true
                                         };
 
     const dispatch = expect.createSpy();
-    const getState = () => ({case: 0, challenge: 0, trial: 0, trials: [{}], authoring: [[{}]] });
+    const getState = () => ({routeSpec: {mission: 0, challenge: 0}, trial: 0, trials: [{}], authoring: [[{}]] });
```
kswenson: Seems like `level` should be included in the `routeSpec`, even though it's apparently not strictly necessary.

test/actions/notifications.js
```diff
@@ -21,9 +21,9 @@ describe('Notification actions', () => {
         {
           message: {
             id: 'ITS.CHALLENGE.INTRO.2',
-            text: 'Ok! Let\'s get to work on Case {{case}} Challenge {{challenge}}.',
+            text: 'Ok! Let\'s get to work on Mission {{mission}} Challenge {{challenge}}.',
             args: {
-              'case': 0,
+              mission: 0,
```
kswenson: Should `level` be included here as well?

test/actions/submit-egg.js
```diff
@@ -22,7 +22,7 @@ describe('submitEggForBasket action', () => {
                             };
 
     const dispatch = expect.createSpy();
-    const getState = () => ({case: 0, challenge: 0, trial: 0, trials: [{}], authoring: [[{}]]});
+    const getState = () => ({mission: 0, challenge: 0, trial: 0, trials: [{}], authoring: [[{}]]});
```
kswenson: Shouldn't this be a `routeSpec`?

test/actions/submit-egg.js
```diff
@@ -50,7 +50,7 @@ describe('submitEggForBasket action', () => {
...
```
kswenson: routeSpec?

src/code/containers/challenge-container-selector.js
```diff
@@ -66,15 +75,14 @@ class ChallengeContainerSelector extends Component {
 function mapStateToProps (state) {
   return {
     authoring: state.authoring,
-    case: state.case,
-    challenge: state.challenge
+    currentRouteSpec: state.routeSpec
```
ekosmin: I did this intentionally to differentiate the `routeSpec` from the `routeParams`, since I kept confusing them. I only changed this one because I didn't want to touch the magically passed `routeParams`.

kswenson: I can see the point, but since it's simply `routeSpec` everywhere else, there's a consistency argument as well. (Which is _not_ to suggest that you should go back and change it everywhere. 😉  I'm just pointing out the issue for future consideration.)

PR #67, opened 2017-02-24
Add screen overlay lines

kswenson:
In my previous PR implementing the background image for Layout C in the Egg Game, I didn't add the overlay accent lines. With this PR, we add the missing overlay accent lines.


src/code/containers/fv-challenge-container.js
```diff
@@ -44,15 +45,18 @@ class FVChallengeContainer extends Component {
...
           scaledTopHudHeight = scaleFactor * topHudHeight,
           scaledImageWidth = scaleFactor * bgImageWidth,
           scaledImageHeight = scaleFactor * bgImageHeight,
           scaledBottomHudHeight = scaleFactor * bottomHudHeight,
           challengeHeight = scaledImageHeight - (scaledTopHudHeight + scaledBottomHudHeight),
-          style = scaledImageHeight ? { height: scaledImageHeight } : {};
+          style = scaledImageHeight
+                    ? { height: scaledImageHeight,
+                        backgroundSize: `auto ${scaledLinesHeight}px, auto ${scaledLinesHeight}px, contain` }
+                    : {};
```
ekosmin: To clarify, is `scaledImageHeight` ever falsy? I can't think of a case where the default `auto auto` style below is used.

kswenson: Good catch! The idea was that the JavaScript scaling overrides should only be applied when necessary. I've fixed the test in a subsequent commit.

PR #69, opened 2017-03-08
Adding information to submitDrake action for ITS

ekosmin:
Here are the changes we discussed for the ITS. I also added a utility function for a common conversion operation. Let me know if I can make any improvements!

### DIFF
--- a/src/code/actions.js
+++ b/src/code/actions.js
@@ -271,6 +271,7 @@ function _submitDrake(targetDrakeIndex, userDrakeIndex, correct, state) {
     species: BioLogica.Species.Drake.name,
     correctPhenotype: targetDrakeOrg.phenotype.characteristics,
     submittedPhenotype: userDrakeOrg.phenotype.characteristics,
+    submittedSex: userDrakeOrg.sex,
     initialAlleles: initialDrakeOrg.alleles,
     selectedAlleles: userDrakeOrg.alleles,
     // Target alleles contains the alleles of the target drake- there may be multiple correct allele strings
--- a/src/code/actions.js
+++ b/src/code/actions.js
@@ -1,5 +1,6 @@
 import actionTypes from './action-types';
 import { ITS_ACTORS, ITS_ACTIONS, ITS_TARGETS } from './its-constants';
+import GeneticsUtils from './utilities/genetics-utils';
 
 export { actionTypes };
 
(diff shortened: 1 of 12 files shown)


ekosmin:
Closing until I can get unit tests running again...

src/code/actions.js
```diff
@@ -257,12 +258,25 @@ function getEarnedCoinString(state) {
...
-    correctPhenotype,
-    submittedPhenotype,
+    species: BioLogica.Species.Drake.name,
+    correctPhenotype: targetDrakeOrg.phenotype.characteristics,
+    submittedPhenotype: userDrakeOrg.phenotype.characteristics,
+    initialAlleles: initialDrakeOrg.alleles,
+    selectedAlleles: userDrakeOrg.alleles,
+    // Target alleles contains the alleles of the target drake- there may be multiple correct allele strings
+    targetAlleles: targetDrakeOrg.alleles,
+    targetSex: targetDrakeOrg.sex,
```
sfentress: I realize that the GUIDE version didn't have `submittedSex`, but I think that must just be an oversight on their part (their little demo game didn't allow the user to change the sex, so they probably forgot). Maybe it should be added?

ekosmin: Yes, I was wondering about that! Added in the newest commit

ekosmin:
Alright, re-opened with unit tests! I was having a lot of trouble getting the `submit-drakes` tests to pass using `calledWith`, so I ended up breaking it down into a number of different equality checks on the dispatched action.

test/actions/submit-drake.js
```diff
@@ -4,30 +4,89 @@ import * as actions from '../../src/code/actions';
...
+      expect(submitDrakeAction.targetSex).toEqual(0);
+      expect(submitDrakeAction.submittedSex).toEqual(0);
+      expect(submitDrakeAction.editableGenes).toEqual("changeableGenes");
+      expect(submitDrakeAction.correct).toEqual(true);
+      expect(submitDrakeAction.incrementMoves).toEqual(false);
+      expect(submitDrakeAction.meta).toEqual({
+        "itsLog": {
+          "actor": "USER",
+          "action": "SUBMITTED",
+          "target": "DRAKE"
```
sfentress: I wonder if this long list of expects (here and for the incorrect action) would be more legible if the objects were compared directly, and doing the phenotype comparison separately. We've done that elsewhere, I don't know if you think it's more legible or more confusing.

The following passes:
```
      expect(submitDrakeAction).toEqual({
        type: types.DRAKE_SUBMITTED,
        species: "Drake",
        correctPhenotype: submitDrakeAction.correctPhenotype,  // we check valid phenotype below
        submittedPhenotype: submitDrakeAction.submittedPhenotype,
        initialAlleles: initialAlleles,
        selectedAlleles: userAlleles,
        targetAlleles: correctAlleles,
        targetSex: 0,
        submittedSex: 0,
        editableGenes: "changeableGenes",
        correct: true,
        incrementMoves: false,
        meta: {
          "itsLog": {
            "actor": "USER",
            "action": "SUBMITTED",
            "target": "DRAKE"
          }
        }
      });
      assertMatchingPhenotype(submitDrakeAction.correctPhenotype, correctCharacteristics);
      assertMatchingPhenotype(submitDrakeAction.submittedPhenotype, userCharacteristics);
      
```

ekosmin:
Thanks for that suggestion @sfentress, still not sure why it doesn't work with everything in the expected action, but this is much cleaner. Also rebased so this is now merging into master

PR #70, opened 2017-03-13
FV Updates

ekosmin:
Adding new stable, chromosomes and gametes


src/code/components/organism.js
```diff
@@ -1,6 +1,8 @@
 import React, {PropTypes} from 'react';
 
-const OrganismView = ({org, id, className="", width=200, flipped=false, style={}, onClick, wrapper }) => {
+const DEFAULT_DRAKE_HEIGHT = 200;
+
+const OrganismView = ({org, id, className="", height, width, flipped=false, style={}, onClick, wrapper }) => {
```
ekosmin: The intention is to be able to create drakes based on height not just width. I've been hitting some issues using two dimensions instead of one, though.

src/code/fv-components/fv-chromosome-image.js
```diff
@@ -0,0 +1,33 @@
+import React, {PropTypes} from 'react';
+import classNames from 'classnames';
+
+const FVChromosomeImageView = ({small=false, bold=false, empty=false, yChromosome=false, animationStyling}) => {
+  let positionStyling = {};
```
ekosmin: Copied and pasted from chromosome image, so some of this code is not used (eg animation styling)

src/code/fv-components/fv-gamete-image.js
```diff
@@ -0,0 +1,119 @@
...
+      <div className="gamete-container" style={containerStyle}>
+        <div style={chromosomeStyle}>
+          { chromosomeImages }
+        </div>
+      </div>
+    </div>
+  );
+};
+
+function getEggImage(scale, fillColor, onClick) {
```
ekosmin: Lots of unused code here...

src/code/fv-components/fv-stable.js
```diff
@@ -0,0 +1,46 @@
...
+
+  let stableDrakeHeight = height * (5/8),
+      stableDrakeViews = orgs.map((org, index) => {
+        return <StableDrakeView org={org} id={idPrefix + index} index={index} key={index}
+                                height={stableDrakeHeight} onClick={handleClick} />;
+      });
+
+  return (
+    <div className="geniblocks fv-stable">
+      <div className="stable-text">{"Stable Count:" + '\xa0\xa0\xa0\xa0\xa0\xa0\xa0\xa0\xa0\xa0\xa0\xa0\xa0\xa0\xa0\xa0\xa0\xa0\xa0\xa0\xa0\xa0\xa0\xa0\xa0\xa0\xa0\xa0\xa0\xa0\xa0\xa0\xa0' + orgs.length + " / 5"}</div>
```
ekosmin: Alright...this is not very clean

src/code/templates/fv-egg-game.js
```diff
@@ -15,20 +15,20 @@ import { motherGametePool, fatherGametePool, gametePoolSelector,
...
 import AnimatedComponentView from '../components/animated-component';
-import ChromosomeImageView from '../components/chromosome-image';
+import FVChromosomeImageView from '../fv-components/fv-chromosome-image';
 import TimerSet from '../utilities/timer-set';
 import t from '../utilities/translate';
 
 // debugging options to shorten animation/randomization for debugging purposes
 // IMPORTANT: remember to set these back to their defaults (e.g. false, false, 0) before committing
-const debugSkipIntroGameteAnimation = false,  // set to true to skip intro gamete animation
+const debugSkipIntroGameteAnimation = true,  // set to true to skip intro gamete animation
```
ekosmin: Not sure we have an intro animation anymore?

src/stylus/fablevision/fv-chromosome.styl
```diff
@@ -0,0 +1,18 @@
+.fv-chromosome-image
+  height: 25px
+  width: 5px
+
+.fv-chromosome-image.small
```
ekosmin: Should be &small, and code below should be moved above

src/stylus/fablevision/stable-drake.styl
```diff
@@ -0,0 +1,16 @@
+.geniblocks .stable-drakes
```
ekosmin: These lines should be combined

src/code/components/organism.js
sfentress: it would be nice not to require that all organism images in the future are square, though.

src/code/fv-components/fv-stable.js
sfentress: Also is this hard-coding a max of five organisms?

src/code/fv-components/stable-drake.js
```diff
@@ -0,0 +1,24 @@
+import React, {PropTypes} from 'react';
+import OrganismView from '../components/organism';
+
+const StableDrakeView = ({org, id, width, height, style, onClick}) => {
```
sfentress: I don't quite understand the need for this view. It's just a single div wrapper around `OrganismView`, right? Couldn't this just be handled in the stable view? i.e. just by adding the wrapper where you currently have `stableDrakeViews = orgs.map...` in `fv-stable.js`?

I realize we already have a few views that just seem to be wrappers (and maybe those can be cleaned up too), but this in particular looks like it just has the one use-case.

src/code/fv-components/fv-chromosome.js
```diff
@@ -0,0 +1,111 @@
...
+import AlleleView from '../components/allele';
+import GeneticsUtils from '../utilities/genetics-utils';
+/**
+ * View of a single chromosome, with optional labels, pulldowns, and embedded alleles.
+ *
+ * Defined EITHER using a Biologica Chromosome object, OR with a Biologica organism,
+ * chromosome name and side.
+ */
+
+const FVChromosomeView = ({chromosome, org, chromosomeName, side, userChangeableGenes = [], visibleGenes = [], hiddenAlleles = [], small = false, editable = true, selected = false, onAlleleChange, onChromosomeSelected, showLabels = true, showAlleles = false, labelsOnRight = true, orgName, displayStyle = {}}) => {
```
sfentress: :+1: on that comment.

src/code/fv-components/fv-stable.js
ekosmin: Changed to be configurable with a default in the new changes

src/code/fv-components/stable-drake.js
ekosmin: This is a good point- since the overlay isn't reusable, I like just wrapping this in the stable instead

src/code/fv-components/stable-counter.js
```diff
@@ -0,0 +1,17 @@
+import React, {PropTypes} from 'react';
+
+const FVStableCounter = ({count, maxCount=5}) => {
```
ekosmin: This component is close to being just a wrapper...but I expect it can be reused in stables for other challenges as well

src/code/components/chromosome.js
```diff
@@ -10,7 +11,12 @@ import GeneticsUtils from '../utilities/genetics-utils';
  * chromosome name and side.
  */
 
-const ChromosomeView = ({chromosome, org, chromosomeName, side, userChangeableGenes = [], visibleGenes = [], hiddenAlleles = [], small = false, editable = true, selected = false, onAlleleChange, onChromosomeSelected, showLabels = true, showAlleles = false, labelsOnRight = true, orgName, displayStyle = {}}) => {
+const ChromosomeView = ({chromosome, org, chromosomeImageClass="ChromosomeImageView", chromosomeName, side, userChangeableGenes = [], visibleGenes = [], hiddenAlleles = [], small = false, editable = true, selected = false, onAlleleChange, onChromosomeSelected, showLabels = true, showAlleles = false, labelsOnRight = true, orgName, displayStyle = {}}) => {
+  const components = {
```
ekosmin: I wonder if it might be better to pass down `basic` or `fablevision` for the image types instead. Those constants could be used for other component selections as well

kswenson: Better would be to pass down the actual class, rather than the name of the class as a string. That way this class doesn't need to know about the set of classes. In fact, I went ahead and made the change to be sure that it would work. I'll add a commit to the PR.

PR #71, opened 2017-03-15
Adding more constants to ITS logs

ekosmin:
Remaining small updates to get the ITS running

### DIFF
--- a/src/code/actions.js
+++ b/src/code/actions.js
@@ -262,9 +262,9 @@ function _submitDrake(targetDrakeIndex, userDrakeIndex, correct, state) {
   const incrementMoves = !correct,
         targetDrakeOrg = GeneticsUtils.convertDrakeToOrg(state.drakes[targetDrakeIndex]),
         userDrakeOrg = GeneticsUtils.convertDrakeToOrg(state.drakes[userDrakeIndex]),
-        initialDrakeOrg = GeneticsUtils.convertDrakeToOrg(state.initialDrakes[userDrakeIndex]),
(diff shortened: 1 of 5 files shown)


src/code/actions.js
```diff
@@ -262,20 +262,18 @@ function _submitDrake(targetDrakeIndex, userDrakeIndex, correct, state) {
   const incrementMoves = !correct,
         targetDrakeOrg = GeneticsUtils.convertDrakeToOrg(state.drakes[targetDrakeIndex]),
         userDrakeOrg = GeneticsUtils.convertDrakeToOrg(state.drakes[userDrakeIndex]),
-        initialDrakeOrg = GeneticsUtils.convertDrakeToOrg(state.initialDrakes[userDrakeIndex]),
+        initialDrakeOrg = state.initialDrakes[userDrakeIndex] ? GeneticsUtils.convertDrakeToOrg(state.initialDrakes[userDrakeIndex]) : null,
         routeSpec = state.routeSpec,
-        editableGenes = state.authoring[routeSpec.level][routeSpec.mission][routeSpec.challenge].userChangeableGenes;
+        editableGenes = state.authoring[routeSpec.level][routeSpec.mission][routeSpec.challenge].visibleGenes.split(", ");
```
kswenson: Seems like the spec should be changed here. If `visibleGenes` is really what's expected, then that's what the field should be called.

ekosmin: Good point, I'll bring this up with Rob during our smoke tests

PR #72, opened 2017-03-15
Layout C challenges respond reasonably to viewport size

kswenson:
Apply top-level CSS scale factor to fit challenge in viewport
- adjust animation computations in FVEggGame to counteract effect of scale
- tweak layout of a number of other objects to work with scaling
- set top-level font size and specify other font-sizes in rem units

### DIFF
--- a/src/code/templates/fv-egg-game.js
+++ b/src/code/templates/fv-egg-game.js
@@ -8,7 +8,7 @@
  */
 
 import React, { Component, PropTypes } from 'react';
-import { assign, clone, cloneDeep, shuffle, range } from 'lodash';
+import { assign, clone, cloneDeep, forIn, shuffle, range } from 'lodash';
 import classNames from 'classnames';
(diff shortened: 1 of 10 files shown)


src/code/templates/fv-egg-game.js
```diff
@@ -93,12 +93,21 @@ function lookupGameteChromosomeDOMElement(sex, chromosomeName) {
   return genomeWrapper.querySelectorAll(".fv-chromosome-image")[chromosomePositions[chromosomeName]];
 }
 
-function findBothElements(sex, name, el){
+function unscaleProperties(obj, scale=1) {
```
ekosmin: We'll probably want this in a utility class at some point, right?

kswenson: Probably, but we can figure out in what form when we need to move beyond this first challenge. It may be that it can get packaged in some coherent animation wrapper component.

PR #74, opened 2017-03-16
ITS improvements

kswenson:
- Queue ITS messages sent before ITS connection established
- Send any queued messages once ITS connection established
- More consistent console.logging of ITS messages
- Dispatch SessionStarted message before any other messages (e.g. default first navigation)
- Dismissing feedback dialog in match drake challenge doesn't dismiss ITS message

### DIFF
--- a/src/resources/authoring/gv2.json
+++ b/src/resources/authoring/gv2.json
@@ -408,7 +408,7 @@
       },
       {
         "comment": "*** FV Test Level 2 ***",
-        "challengeId": "gamete-1target-simple",
+        "challengeId": "egg-breeding-basic-match-visible",
         "container": "FVContainer",
         "template": "FVEggGame",
         "challengeType": "match-target",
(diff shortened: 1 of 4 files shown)


src/resources/authoring/gv2.json
```diff
@@ -408,7 +408,7 @@
       },
       {
         "comment": "*** FV Test Level 2 ***",
-        "challengeId": "gamete-1target-simple",
+        "challengeId": "egg-breeding-basic-match-visible",
```
ekosmin: `gamete-1target-simple` is the ID that Trudi came up with, while `egg-breeding-basic-match-visible` is what I was using before I had anything from her. The ITS server can handle `gamete-1target-simple`, so I think we should stick with that

kswenson: ok, I was hypothesizing that that might have been why the ITS wasn't working, but apparently that was a red herring.

PR #76, opened 2017-03-17
Layout C Improvements - Breed Button Area

kswenson:
Add BreedButtonAreaView
- renders FableVision background decoration
- renders BreedButtonView
- renders gamete (egg and sperm) FableVision images appropriately, including tail of sperm
  - (still uses pre-FableVision chromosome images inside the gametes)
- renders egg/child drake once fertilization is complete
- renders Submit/Reset buttons once fertilization is complete

Code for rendering all of the above removed from FVEggGame template

Other layout improvements
- adjusts position of parent drake views
- adjusts egg/sperm breeding animation motion


kswenson:
Note: Rebased to master to capture Eli's chromosome label styling changes

src/stylus/fablevision/fv-gamete.styl
```diff
@@ -8,10 +8,13 @@
   align-items: center
 
 .sperm .gamete-container
-  background: url('../resources/fablevision/layout-c/sperm_head.png')
-  background-size: contain
+  background-image: url('../resources/fablevision/layout-c/sperm_head.png'),
+                    url('../resources/fablevision/layout-c/sperm_tail.png')
+  // pin the tail on the sperm
```
ekosmin: Huzzah!

src/code/fv-components/fv-gamete-image.js
```diff
@@ -28,18 +28,19 @@ const FVGameteImageView = ({chromosomes=[], id, className, style, displayStyle})
     chromosomeImages.push(
       <ChromosomeImageView
         key={i}
-        empty={!chromosome}
-        width={6 * scale}
-        height={(chromosome && chromosome.side === "y" ? 15 : 20) * scale}
-        split={8 * scale}
+        empty={false}
```
ekosmin: This is the only confusing bit- is the idea that empty sets the color to white, and we always want to use the colors passed below? Not that this is worth messing with now, but a passed color should probably override any flags, right?

kswenson: These are the old SVG chromosomes, so we're not going to be using them for very long in any case. Empty means show the outline but don't fill the interior. Previously, we distinguished selection by having a fillColor versus being empty. To look more like the FableVision art in the short term, I simply have it always fill (i.e. never empty) with either the dark or light blue chromosome color. It will all go away when we start using the FableVision images, but this seemed like a simple quick-fix to improve the look and could be good enough for the classroom test.

src/stylus/fablevision/fv-gamete.styl
kswenson: It actually took a non-trivial amount of time to line the images up properly. I thought about switching to the single whole-sperm image, but it has a white background (not transparent) so it wouldn't have worked without getting new art from FableVision.

ekosmin: I wen

PR #77, opened 2017-03-17
Updating chromosome layouts to FV style

ekosmin:
Simple fix to update chromosome labels to static text, switching on the `ChromosomeImageClass`

### DIFF
--- a/src/code/components/chromosome.js
+++ b/src/code/components/chromosome.js
@@ -27,13 +27,16 @@ const ChromosomeView = ({chromosome, org, ChromosomeImageClass=ChromosomeImageVi
 
     if (showLabels) {
       let labels = visibleAlleles.map(a => {
-        return (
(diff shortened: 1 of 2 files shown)


src/code/components/chromosome.js
```diff
@@ -27,13 +27,16 @@ const ChromosomeView = ({chromosome, org, ChromosomeImageClass=ChromosomeImageVi
 
     if (showLabels) {
       let labels = visibleAlleles.map(a => {
-        return (
+        if (ChromosomeImageClass === ChromosomeImageView) {
```
kswenson: There's probably a better abstraction lurking here. Perhaps an `FVGeneLabelView` and then have the client pass in the `LabelClass`. I don't think it's worth the time to figure that out before the classroom test so I'll merge it as is for now as it's clearly an improvement visually.

PR #78, opened 2017-03-17
Reset & Submit buttons use new FableVision art

kswenson:
- move responsibility for Reset/Submit buttons out of BreedButtonArea and back into FVEggGame
- move layout of buttons to be at top of page rather than next to egg/drake
- increase size of user drake
- use <a> tag for buttons to show clicked/unclicked state


kswenson:
Rebased to master to incorporate Eli's background lines

PR #80, opened 2017-03-18
Windows support

kswenson:
- add polyfills for IE11 support
- add text adjustments for all Windows browsers

All Windows browsers tested (Chrome, Firefox, IE11) showed the same text offsets, suggesting that it may be a font-metrics issue rather than a browser issue. We "fix" it for the time being by user agent-sniffing for Windows and adding the `os-windows` class to the `<body>` tag if appropriate. Then CSS is used on a case-by-case basis to adjust the text where appropriate.

Note: after review, let me merge so I can rebase appropriately.

### DIFF
--- a/src/code/containers/modal-message-container.js
+++ b/src/code/containers/modal-message-container.js
@@ -2,7 +2,8 @@ import { connect } from 'react-redux';
 import ModalAlert from '../components/modal-alert';
 import * as actions from '../actions';
 import { resetGametes } from '../modules/gametes';
-Object.assign(actions, { resetGametes });
+// merge with the other actions
+actions.resetGametes = resetGametes;
 
 function mapStateToProps (state) {
   let props = state.modalDialog;
--- a/src/stylus/fablevision/parent-drake.styl
+++ b/src/stylus/fablevision/parent-drake.styl
@@ -33,3 +33,8 @@
     margin-top: 10px
   .organism
     margin-top: -8px
+
+.os-windows
+  .geniblocks.parentDrake
+    .parentTitle
+      margin-top: 1rem
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "axios": "^0.15.3",
     "babel-polyfill": "^6.20.0",
     "classnames": "^2.2.5",
+    "es6-object-assign": "^1.0.3",
     "lodash": "^4.17.3",
     "mustache": "^2.3.0",
     "react": "^15.4.1",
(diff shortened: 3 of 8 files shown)


src/code/containers/modal-message-container.js
```diff
@@ -2,7 +2,8 @@ import { connect } from 'react-redux';
 import ModalAlert from '../components/modal-alert';
 import * as actions from '../actions';
 import { resetGametes } from '../modules/gametes';
-Object.assign(actions, { resetGametes });
+// merge with the other actions
+actions.resetGametes = resetGametes;
```
ekosmin: Is this change necessary, now that `Object.assign` works on all browsers?

src/stylus/fablevision/parent-drake.styl
```diff
@@ -33,3 +33,8 @@
     margin-top: 10px
   .organism
     margin-top: -8px
+
+.os-windows
+  .geniblocks.parentDrake
+    .parentTitle
+      margin-top: 1rem
```
ekosmin: Were you able to find any consistent way to determine how much the Windows offset needs to be? Checking for every text component seems tedious

src/code/containers/modal-message-container.js
kswenson: Still not in IE11.

src/stylus/fablevision/parent-drake.styl
kswenson: Not yet, but I haven't looked very hard. I think manually tweaking the handful we care about for the classroom test makes sense, and then we can take the time to look for a more general solution after the classroom test.

src/code/containers/modal-message-container.js
ekosmin: My impression was that the purpose of the polyfill was to make it work in IE11, but I must misunderstand. We can catch up about it offline

src/stylus/fablevision/parent-drake.styl
ekosmin: Makes sense!

src/code/containers/modal-message-container.js
kswenson: Ah, the reason I eliminated the use of Object.assign in this particular case was because I didn't like the idea of having this module be responsible for loading the polyfill rather than having it handled centrally. The only reason that was necessary was because this module was being loaded by the browser before the modules that loaded the polyfill. But rather than figure that out, it was easier to eliminate the use of Object.assign() in this case.

PR #81, opened 2017-03-18
Add trial counter to bottom HUD

kswenson:
- new TrialCountView shows trial count text inside FableVision trial counter art
- new BottomHUDView places TrialCountView appropriately

Note: after review, let me merge so I can rebase appropriately.


src/code/fv-components/bottom-hud.js
```diff
@@ -0,0 +1,20 @@
+import React, {PropTypes} from 'react';
+import TrialCountView from './trial-count';
+
+const BottomHUDView = ({id, className, trial, trialCount}) => {
+
+  return (
```
ekosmin: I suppose this is worth a component because the HUD will be expanding a lot, right?

src/stylus/fablevision/trial-count.styl
```diff
@@ -0,0 +1,25 @@
+.geniblocks.trial-count
+  width: 158px
+  height: 107px
+  background-image: url('../resources/fablevision/huds/Trial_screen.png')
+  background-size: 100% 100%
```
ekosmin: Interesting, I've been using contain...either solution requires us looking up the size of the image, which doesn't feel great. Is there a better way to do this?

src/stylus/fablevision/trial-count.styl
```diff
@@ -0,0 +1,25 @@
...
+
+  .text-area
```
ekosmin: I'm still trying to figure out how to name my CSS classes. Would it be better to name this something more specific like `trial-text` to give the impression that it's not a more generic text area? Or is this preferred so we could make more global changes down the line?

src/code/fv-components/bottom-hud.js
kswenson: Yes, there will likely be some layout logic, e.g. hiding/showing characters/messages, etc.

src/stylus/fablevision/trial-count.styl
kswenson: `contain` means maintain aspect ratio, leaving empty space if necessary. `100% 100%` means fill the area, stretching the image if necessary. In this case, I'm willing to stretch the image because I could imagine needing a bit more horizontal space and stretching the image probably wouldn't be too bad in this case, which could mean not having to go back to FableVision for a larger image. I'm not aware of any way to have the element auto-sized to the size of the background, given that the size of the background is generally specified relative to the size of the element.

src/stylus/fablevision/trial-count.styl
kswenson: I agree it's not always obvious. In this case I was imagining that there might be other HUD text areas for which similar rules might apply in other HUD elements. Perhaps `hud-text-area` would make more sense? I think we can leave it for now until we see whether others arise.

PR #83, opened 2017-03-19
Adding white lines to chromosomes and updating gamete chromosomes

ekosmin:


kswenson:
The approach of having a single chromosome image with two stripes will work for the classroom test, but ultimately since there can be anywhere from 0 (as in X/Y) to four or five genes per chromosome, we'll need a more dynamic approach to drawing the lines on the chromosomes.

ekosmin:
^ Certainly true- genes also have different positions on chromosomes, so we'll need to be able to dynamically read those. Like you said, this is just a temporary solution for classroom testing

PR #84, opened 2017-03-19
Completing last challenge...

kswenson:
- gives a "Mission Accomplished!" message
- offers options to retry the current challenge or mission
- no longer wraps around to the first challenge [#142003363]

### DIFF
--- a/src/code/actions.js
+++ b/src/code/actions.js
@@ -58,6 +58,13 @@ export function retryCurrentChallenge() {
   };
 }
 
+export function retryCurrentMission() {
+  return (dispatch, getState) => {
+    const { level, mission } = getState().routeSpec;
+    dispatch(navigateToChallenge({ level, mission, challenge: 0 }));
+  };
+}
+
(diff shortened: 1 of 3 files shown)


src/code/actions.js
```diff
@@ -290,25 +297,55 @@ function _submitDrake(targetDrakeIndex, userDrakeIndex, correct, state) {
...
+          missions = levels[state.routeSpec.level],
+          missionCount = missions.length,
+          challenges = missions[state.routeSpec.mission],
+          challengeCount = challenges.length,
+          trials = state.trials,
+          trialCount = trials.length;
     let challengeComplete = false,
-        missionComplete = false;
+        missionComplete = false,
+        // levelComplete = false,
```
ekosmin: Probably worth getting rid of unused variables

kswenson: I left this one in because I anticipate that we will want to give a unique end-of-level message at some point.

PR #85, opened 2017-03-19
Add location indicator to top HUD [#142012541]

kswenson:
Note that more recent layout images from FableVision show a different location indicator graphic, but we don't have access to that updated image at this time.

Also,
- use absolute positioning for HUD layout
- increase size of stable count text
- clip target drake gizmo to edge of layout


src/code/fv-components/location-indicator.js
```diff
@@ -0,0 +1,18 @@
+import React, {PropTypes} from 'react';
```
ekosmin: Is it worth making this its own component? Sam made a comment earlier that one of my component's was just text in a div and probably not worth breaking out, so I'm trying to figure out best practices.

src/stylus/fablevision/location-indicator.styl
```diff
@@ -0,0 +1,25 @@
...
+  .hud-text-area
+    position: relative
+    left: 90px
+    top: 16px
+    width: 770px
+    display: flex
+    flex-direction: column
+    justify-content: center
+    align-items: center
+    font-family: 'Yanone Kaffeesatz'
```
ekosmin: At some point should we move the font selection to a higher class rather than putting it in each?

src/stylus/fablevision/target-drake.styl
```diff
@@ -1,24 +1,23 @@
 .geniblocks.target-drake-container
   display: flex
   flex-direction: row
-  height: 588px
+  height: 272px
   width: 520px
   background: url('../resources/fablevision/layout-c/target_drake_gizmo.png')
   background-repeat: no-repeat
-  background-size: contain
+  background-size: cover
```
ekosmin: Clever change! I noticed my component bled off the page before.

src/stylus/fablevision/location-indicator.styl
kswenson: Yes, I think once we get past this classroom test we should give some thought to better organization of the CSS. I think it's useful to have a few instances first so we have a better idea of what the general case looks like.

src/code/fv-components/location-indicator.js
kswenson: There's certainly a gray area and different engineers will decide differently in different circumstances. One of the criteria I use when thinking about it is how well does the component stand on its own -- for instance, how likely is it to be useful in another context? In the case of the stable drakes, the effect was intended to be specific to the stable, so there's an argument for having the stable define the effect. In this case, I feel like it's the HUD's job to lay out the components on the HUD, but the components themselves should be self-contained.

ekosmin: That makes sense! Thanks for breaking down your thought process.

PR #86, opened 2017-03-20
Enable CC logging [#142012533]

kswenson:
- logging enabled for non-development use by default
- logging disabled for development by default
- logging can be enabled for development by URL parameter

Note: I don't have access to the log manager, so I haven't been able to check what's being logged at the other end, but what's being sent from the client looks reasonable.

### DIFF
--- a/src/code/middleware/gv-log.js
+++ b/src/code/middleware/gv-log.js
@@ -1,7 +1,15 @@
 import actionTypes from '../action-types';
 import templates from '../templates';
+import urlParams from '../utilities/url-params';
 
-//const logManagerUrl  = '//cc-log-manager.herokuapp.com/api/logs';
+function isLocalHostUrl() {
+  const host = window.location.hostname;
+  return (host.indexOf('localhost') >= 0) || (host.indexOf('127.0.0.1') >= 0);
+}
+
+const logManagerUrl  = '//cc-log-manager.herokuapp.com/api/logs',
+      // disable logging during development unless explicitly enabled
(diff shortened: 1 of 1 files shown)


sfentress:
This looks good, and is logging.

The one thing I would add, if you have time, is the `activity` property, which is a top-level LogManager property for designating a particular "activity" in a project, which in this case I think would map to challenge.

I know that eventually we want to have unique challengeIds for each challenge, so this would probably be the best string to use. We could have a hybrid for the moment, as some challenges don't have names, but maybe it's best to just use the name if we have it, and not if we don't?

Hmmm... although looking at the state now, I see that `challengeId` is not visible, except by drilling down into the authoring. I guess you're already grabbing this string to use for ITS, so it wouldn't be too bad to use it here as well. I'm guessing eventually this string might be found in the routeSpec?

As for accessing the Log Manager, if you and @ekosmin want to make accounts at http://cc-log-manager.herokuapp.com/users/sign_up I can give you access.

sfentress:
To clarify, the suggested change would be something like

```
function createLogEntry(loggingMetadata, action, nextState){
    ...
    let routeSpec = nextState.routeSpec,
        activity  = nextState.authoring[routeSpec.level][routeSpec.mission][routeSpec.challenge].challengeId;

    const message =
    {
      application: loggingMetadata.applicationName,
      activity: activity,
````

kswenson:
Following Sam's suggestion, added `activity` to the log messages.

sfentress:
Looks great. :+1:

PR #87, opened 2017-03-20
Final(?) tweaks

kswenson:
- Adjust color of disabled breed button text [##142003501]
- Reset gametes on new challenge

### DIFF
--- a/src/code/templates/fv-egg-game.js
+++ b/src/code/templates/fv-egg-game.js
@@ -822,7 +822,7 @@ export default class FVEggGame extends Component {
                             showHatchAnimation: showUserDrake,
                             clearAnimatedComponents: true });
       this.autoSelectSingletonGametes(nextGametes);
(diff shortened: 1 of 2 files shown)


src/code/templates/fv-egg-game.js
```diff
@@ -822,7 +822,7 @@ export default class FVEggGame extends Component {
                             showHatchAnimation: showUserDrake,
                             clearAnimatedComponents: true });
       this.autoSelectSingletonGametes(nextGametes);
-      if (newTrialInChallenge && onResetGametes) {
+      if ((newChallenge || newTrialInChallenge) && onResetGametes) {
```
ekosmin: Nice, this has been an outstanding issue for a while. Should we do this for new levels and missions as well?

kswenson: If the level/mission changes, the `newChallenge` is set to true.

PR #88, opened 2017-03-20
Fix animations when document is scrolled

kswenson:
- unscaling code must take scroll position into account

Note: although I'm tempted to try to get this in before tomorrow's classroom test, I think prudence suggests that we hold off on any more deployments unless a true showstopper is encountered.

### DIFF
--- a/src/code/utilities/unscale-properties.js
+++ b/src/code/utilities/unscale-properties.js
@@ -1,10 +1,32 @@
 import { forIn } from 'lodash';
 
+export function unscaleXPos(value, scale) {
+  // account for horizontal scroll position
+  return (value + window.pageXOffset) / scale;
+}
+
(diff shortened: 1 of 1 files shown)


src/code/utilities/unscale-properties.js
```diff
@@ -1,10 +1,32 @@
...
+  // account for horizontal scroll position
+  return (value + window.pageXOffset) / scale;
+}
+
+export function unscaleYPos(value, scale) {
+  // account for vertical scroll position
+  return (value + window.pageYOffset) / scale;
+}
+
+export function unscaleExtent(value, scale) {
```
ekosmin: Are extents things like width and height?

kswenson: Yes, extents/lengths aren't affected by scroll position.

PR #90, opened 2017-03-21
Fix previous challenges

kswenson:
Fix a handful of issues in which changes for the new layout adversely affected old challenges. Some of these are layout issues, some are functionality.

- Redo some layout changes that adversely affected previous challenges in ways that don't have the adverse effects.
- Like regular guide messages, guide alert messages shouldn't dismiss dialogs before users have had a chance to read them
- Add null test for `visibleGenes` before dereferencing it
- Lock Biologia.js version at 0.0.2 to force usage of old allele strings
- User always gets gold coin for first playground level

Note: Currently depends on PR#89. This PR should be rebased once that one has been merged.
PR#89 has been merged and this PR has been rebased.

### DIFF
--- a/src/code/templates/genome-playground.js
+++ b/src/code/templates/genome-playground.js
@@ -18,13 +18,15 @@ export default class GenomeContainer extends Component {
     let penView = <div className='columns bottom'>
                     <PenView orgs={ keptDrakes } width={500} columns={5} rows={1} tightenRows={20}/>
                   </div>;
-    
+
+    // we don't increment the score, so users always get top mark from playground
+    const kNoScoreIncrement = false;
     const handleAlleleChange = function(chrom, side, prevAllele, newAllele) {
-      onChromosomeAlleleChange(0, chrom, side, prevAllele, newAllele);
+      onChromosomeAlleleChange(0, chrom, side, prevAllele, newAllele, kNoScoreIncrement);
     };
 
     const handleSexChange = function(newSex) {
-      onSexChange(0, newSex);
+      onSexChange(0, newSex, kNoScoreIncrement);
     };
 
     const handleSubmit = function() {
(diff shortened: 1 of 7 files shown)


src/code/templates/genome-playground.js
```diff
@@ -18,13 +18,15 @@ export default class GenomeContainer extends Component {
     let penView = <div className='columns bottom'>
                     <PenView orgs={ keptDrakes } width={500} columns={5} rows={1} tightenRows={20}/>
                   </div>;
-    
+
+    // we don't increment the score, so users always get top mark from playground
+    const kNoScoreIncrement = false;
```
ekosmin: I'm confused by this variable name- why choose something other than `incrementMoves`?

kswenson: It's a constant intended to be passed as an argument to a function which accepts a boolean argument named `incrementMoves`. If the constant/variable were named `incrementMoves` as well, then the function call would look like:
```
onChromosomeAlleleChange(..., incrementMoves);
```
which could be read as a request to increment the moves rather than to _not_ increment the moves, as is actually the case. Without a constant, the code looks like:
```
onChromosomeAlleleChange(..., false);
```
which is uninterpretable without scrambling to look up the function definition. Passing a semantically meaningful constant to the argument named `incrementMoves` is intended to make the intent of the code more clear. Whether or not the precise name of the constant is as semantically meaningful as I originally could be a matter for further debate.

ekosmin: Got it, perhaps I'm not familiar with the notation style, is it related to this? http://stackoverflow.com/questions/28762564/why-do-variable-names-start-with-the-letter-k

The way I read it makes it seem like `noScoreIncrement = false`, which means `scoreIncrement = true` by a double negative... is a constant `doesPlaygroundIncrementMoves` any clearer? I agree the code is confusing without any named constant- is it in general best practice to name parameters if their use is ambiguous?

These are very small nits, thank you for entertaining them! More than anything else I'm trying to figure out the best style to use myself moving forward.

PR #91, opened 2017-03-21
Add baseUser URL parameter [#142141187]

kswenson:
- `baseUser` defaults to "gv2-user" if URL parameter is not specified
- `userName` = "{baseUser}-{partialSessionID}"
- `userName` is constructed in app.js and passed to CC logger and ITS via `loggingMetadata`

Note: Currently depends on PR#89. This PR should be rebased once that one has been merged.
PR#89 has been merged and this PR has been rebased.

### DIFF
--- a/src/code/app.js
+++ b/src/code/app.js
@@ -43,7 +43,7 @@ window.GV2Authoring = convertAuthoring(authoring);
 
 // TODO: session ID and application name could be passed in via a container
 // use placeholder ID for duration of session and hard-coded name for now.
-const loggingMetadata = {
+let loggingMetadata = {
   applicationName: "GeniStarDev"
 };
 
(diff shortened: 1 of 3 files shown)


sfentress:
I think this looks fine. It will be easier to tell when the other PR's are merged in, since this one includes those. (I'm leaving merging to you, since I don't want to mess up any hotfix deployments you might need to do during the run.)

Is the username url parameter going to be the same for a whole group of students, or are they each going to have their own one? I assume the former, as otherwise I'd think we'd be passing that through as is, instead of tacking on the session id.

kswenson:
I'm assuming this will be used for a group of student as you surmised, so that for a classroom test such as the one today, `baseUser=normsStudent` would separate sessions related to the classroom test from all other sessions. There are plans to support individual logins down the road, but not for a while yet.

PR #94, opened 2017-03-22
Authoring restructure

ekosmin:
Updated the authoring doc per this story:

https://www.pivotaltracker.com/story/show/142096149

The structure is now broken into level definitions and a level hierarchy. Additionally, at the suggestion of @kswenson, I use objects instead of strings in the level hierarchy so that we could move application-specific logic into the hierarchy and make the levels more reusable. The idea is that we could move more applications into a single authoring document, but I didn't have the opportunity to do that, and want to prioritize other tasks.

The downside of using objects instead of strings is that it makes the authoring document a little more challenging for non-engineers to modify. What do you think?

### DIFF
--- a/src/code/actions.js
+++ b/src/code/actions.js
@@ -88,14 +88,16 @@ export function navigateToNextChallenge() {
         nextLevel = routeSpec.level,
         nextMission = routeSpec.mission,
         nextChallenge = currentChallenge+1,
-        challengeCountInMission = authoring.levelHierarchy[currentLevel][currentMission].length;
+        challengeCountInMission = AuthoringUtils.getChallengeCount(authoring, currentLevel, currentMission);
     if (challengeCountInMission <= nextChallenge) {
+      let missionCountInLevel = AuthoringUtils.getMissionCount(authoring, currentLevel);
       // there are not enough challenges...if the next mission exists, navigate to it
-      if (authoring.levelHierarchy[currentLevel][currentMission+1]) {
+      if (currentMission + 1 < missionCountInLevel) {
         nextMission++;
       } else {
+        let levelCount = AuthoringUtils.getLevelCount(authoring);
         // otherwise, check if the next level exists
-        if (authoring.levelHierarchy[currentLevel+1]) {
+        if (currentLevel + 1 < levelCount) {
           nextLevel++;
         } else {
           // if no next level exists, loop around
@@ -147,12 +149,12 @@ function restrictToIntegerRange(value, minValue, maxValue) {
 export function navigateToCurrentRoute(routeSpec) {
   const {level, mission, challenge} = routeSpec;
   return (dispatch, getState) => {
-    const levels = getState().authoring.levelHierarchy,
-          levelCount = levels.length,
+    const authoring = getState().authoring,
+          levelCount = AuthoringUtils.getLevelCount(authoring),
           nextLevel = restrictToIntegerRange(level, 0, levelCount - 1),
-          missionCount = levels[nextLevel].length,
+          missionCount = AuthoringUtils.getMissionCount(authoring, nextLevel),
           nextMission = restrictToIntegerRange(mission, 0, missionCount - 1),
-          challengeCount = levels[nextLevel][nextMission].length,
+          challengeCount = AuthoringUtils.getChallengeCount(authoring, nextLevel, nextMission),
           nextChallenge = restrictToIntegerRange(challenge, 0, challengeCount - 1),
           routeChangeRequired = (level !== nextLevel) || (mission !== nextMission) || (challenge !== nextChallenge);
     // TODO: Ideally, route changes would be handled in their own module (see routing.js)
@@ -306,12 +308,10 @@ export function submitDrake(targetDrakeIndex, userDrakeIndex, correct, incorrect
     const state = getState();
     dispatch(_submitDrake(targetDrakeIndex, userDrakeIndex, correct, state));
 
-    const levels = state.authoring.levelHierarchy,
-          levelCount = levels.length,
-          missions = levels[state.routeSpec.level],
-          missionCount = missions.length,
-          challenges = missions[state.routeSpec.mission],
-          challengeCount = challenges.length,
+    const authoring = state.authoring,
+          levelCount = AuthoringUtils.getLevelCount(authoring),
+          missionCount = AuthoringUtils.getMissionCount(authoring, state.routeSpec.level),
+          challengeCount = AuthoringUtils.getChallengeCount(authoring, state.routeSpec.level, state.routeSpec.mission),
           trials = state.trials,
           trialCount = trials.length;
     let challengeComplete = false,
@@ -495,7 +495,7 @@ export function showCompleteChallengeDialog() {
   return (dispatch, getState) => {
     const state = getState();
 
-    const missionComplete = (state.authoring.levelHierarchy[state.routeSpec.level][state.routeSpec.mission].length <= state.routeSpec.challenge + 1);
+    const missionComplete = (AuthoringUtils.getChallengeCount(state.authoring, state.routeSpec.level, state.routeSpec.mission) <= state.routeSpec.challenge + 1);
 
     let dialog = {};
 
--- a/src/code/actions.js
+++ b/src/code/actions.js
@@ -1,6 +1,7 @@
 import actionTypes from './action-types';
 import { ITS_ACTORS, ITS_ACTIONS, ITS_TARGETS } from './its-constants';
 import GeneticsUtils from './utilities/genetics-utils';
+import AuthoringUtils from './utilities/authoring-utils';
 
 export { actionTypes };
 
@@ -87,14 +88,14 @@ export function navigateToNextChallenge() {
         nextLevel = routeSpec.level,
         nextMission = routeSpec.mission,
         nextChallenge = currentChallenge+1,
-        challengeCountInMission = authoring[currentLevel][currentMission].length;
+        challengeCountInMission = authoring.levelHierarchy[currentLevel][currentMission].length;
     if (challengeCountInMission <= nextChallenge) {
       // there are not enough challenges...if the next mission exists, navigate to it
-      if (authoring[currentLevel][currentMission+1]) {
+      if (authoring.levelHierarchy[currentLevel][currentMission+1]) {
         nextMission++;
       } else {
         // otherwise, check if the next level exists
-        if (authoring[currentLevel+1]) {
+        if (authoring.levelHierarchy[currentLevel+1]) {
           nextLevel++;
         } else {
           // if no next level exists, loop around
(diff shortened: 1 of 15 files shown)


src/resources/authoring/gv2.json
```diff
@@ -1,67 +1,60 @@
...
-          "alleles": "",
-          "sex": 1
-        }],
-        "userChangeableGenes": "wings, forelimbs, hindlimbs",
-        "linkedGenes": {
-          "drakes": [0, 1],
-          "genes": "metallic, tail, horns, color, black, dilute, armor, nose, bogbreath"
-        }
+{
+  "definitions": {
```
ekosmin: I remember we discussed a different name for this portion of the document @kswenson, do you remember what it was?

src/resources/authoring/gv2.json
```diff
@@ -426,6 +398,57 @@
         },
         "visibleGenes": "metallic, wings, forelimbs, hindlimbs"
       }
+  },
+
+
+  "levelHierarchy": [
+    // Level 1
```
sfentress: These comments make the file invalid JSON. Depending on how `require` pulls it in, it may make not cause problems now, but it may bite us if we do something like push this into Firebase.

src/resources/authoring/gv2.json
sfentress: Definitions seems a little vague to me, personally. Definitions of what? How about just "challenges"? Or is the worry that that name may not be correct in the future?

src/code/actions.js
```diff
@@ -271,7 +272,9 @@ function _submitDrake(targetDrakeIndex, userDrakeIndex, correct, state) {
         userDrakeOrg = GeneticsUtils.convertDrakeToOrg(state.drakes[userDrakeIndex]),
         initialDrakeOrg = state.initialDrakes[userDrakeIndex] ? GeneticsUtils.convertDrakeToOrg(state.initialDrakes[userDrakeIndex]) : null,
         routeSpec = state.routeSpec,
-        editableGenes = state.authoring[routeSpec.level][routeSpec.mission][routeSpec.challenge].visibleGenes.split(", ");
+        challengeDefinition = AuthoringUtils.getChallengeDefinition(state.authoring, routeSpec),
+        editableGeneString = challengeDefinition.visibleGenes || challengeDefinition.userChangeableGenes,
```
sfentress: Why are the visible genes the default ones we're submitting for "editableGenes"? If a challenge has `editableGenes: "tail", visibleGenes: "wings"`, then this means that the user can change "tail", and sees "wings" as a static, non-editable gene. With this, we'd just be submitting "wings".

src/resources/authoring/gv2.json
kswenson: I was thinking `challenges`, since it really is just a list of challenges.

src/resources/authoring/gv2.json
kswenson: We should have valid JSON. It's potentially another argument for having objects which could then have additional fields for `comments`, `description`, etc. It could look something like:
```
{
  challenges: {
    ...
  },
  application: {
    "name": "...",
    "description": "...",
    "levels": [
      {
        "name": "Level 1",
        "description": "...",
        "missions": [
          {
            "name": "Mission 1",
            "description": "...",
            "challenges": [
              {
                "name": "Challenge 1",
                "location": "Hatchery",
                "challengeId": "someUniqueChallengeID"
              },
              {
                "name": "Challenge 2",
                "location": "Hatchery",
                "challengeId": "anotherUniqueChallengeID"
              }
            ]
          },
          {
            "name": "Mission 2",
            "description": "...",
            "challenges": [
              {
                "name": "Challenge 1",
                "location": "Hatchery",
                "challengeId": "yetAnotherUniqueChallengeID"
              },
              {
                "name": "Challenge 2",
                "location": "Hatchery",
                "challengeId": "stillAnotherUniqueChallengeID"
              }
            ]
          }
        ]
      },
      {
        "name": "Level 2",
        "missions": []
      }
    ]
  }
}
```

src/code/actions.js
```diff
@@ -1,6 +1,7 @@
 import actionTypes from './action-types';
 import { ITS_ACTORS, ITS_ACTIONS, ITS_TARGETS } from './its-constants';
 import GeneticsUtils from './utilities/genetics-utils';
+import AuthoringUtils from './utilities/authoring-utils';
```
kswenson: I like the idea of collection utility functions for accessing the authoring structure into a single module. I would even go further and have as much of the access as possible mediated through such functions, so that the authoring structure could be changed with minimal effect on other clients.

src/code/actions.js
ekosmin: Good point, I should swap the ordering here

kswenson: This originally came out of a request from Rob shortly before the classroom test. We need to circle back with him to figure out what the ITS is expecting and how it's interpreting these fields. We shouldn't be muddying the waters here. Also, @ekosmin , there will be a conflict here as I also addressed a crash that occurred in older challenges that didn't have the requisite fields. You should rebase to master before making further changes here.

src/code/actions.js
```diff
@@ -146,12 +149,12 @@ function restrictToIntegerRange(value, minValue, maxValue) {
...
-    const { authoring: levels } = getState(),
-          levelCount = levels.length,
+    const authoring = getState().authoring,
+          levelCount = AuthoringUtils.getLevelCount(authoring),
           nextLevel = restrictToIntegerRange(level, 0, levelCount - 1),
-          missionCount = levels[nextLevel].length,
+          missionCount = AuthoringUtils.getMissionCount(authoring, nextLevel),
           nextMission = restrictToIntegerRange(mission, 0, missionCount - 1),
-          challengeCount = levels[nextLevel][nextMission].length,
+          challengeCount = AuthoringUtils.getChallengeCount(authoring, nextLevel, nextMission);
```
kswenson: `;` should be `,`

ekosmin:
This code is ready for another round of review! The first big change is that now all references to internal `challenges` and `levelHierarchy` variables in the authoring doc are contained in the `authoring-utils`. The other main difference is that you can now address levels by challenge ID like so: http://127.0.0.1:8080/gv2/?challengeId=basicsIntroHidden#/. Challenges progress normally from there.

src/code/actions.js
ekosmin: Sorry, I pushed an unfinished version, I should have unassigned in the meantime. It works now!

src/code/containers/challenge-container-selector.js
```diff
@@ -43,27 +44,33 @@ class ChallengeContainerSelector extends Component {
   }
 
   componentWillMount() {
-    const { routeParams, navigateToCurrentRoute, navigateToChallenge } = this.props;
-    if (hasChangedRouteParams(this.props)) {
+    const { navigateToCurrentRoute, navigateToChallenge, authoring } = this.props;
+    // the URL's challengeId is only used for initial routing, so prioritize the numeric route params
```
ekosmin: This is the main part I have questions about- I have numerical paths take priority over written challenge IDs because it required the least effort on my part to update the URL. Does that seem reasonable for now?

sfentress: I don't quite understand having the challengeId be in the urlParams and the routeParams be in the hash. It seems to me that as the user moves forward, there will be a mismatch between challengeId and the routeSpec, which will be confusing. After finishing `?challengeId=basicsIntroVisible` I will go to `?challengeId=basicsIntroVisible#/2/1/2`, which doesn't seem right.

I think what would make more sense for the user would be to be able to go to `#/basicsIntroVisible` (which is also easier to remember, because you don't need the `challengeId=`), have that load up, and then if you continue then your url changes to `#/2/1/2`.

It looks like this wouldn't be too hard. I tested the following, but I may have missed something:

Change app.js to include the new route definition:
```
<Router history={history}>
  <Route path="/(:level/:mission/:challenge)" component={ChallengeContainerSelector} />
  <Route path="/(:challengeId)" component={ChallengeContainerSelector} />
</Router>
```
Change challenge-container-selector.js to look for the challengeId in the routeParams:
```
componentWillMount() {
    const { navigateToCurrentRoute, navigateToChallenge, authoring } = this.props;
    let routeParams = this.props.routeParams;
    if (routeParams.challengeId) {
      routeParams = AuthoringUtils.challengeIdToRouteParams(authoring, routeParams.challengeId);
    }
    if (hasChangedRouteParams(this.props.currentRouteSpec, routeParams)) {
      navigateToCurrentRoute({level: routeParams.level-1, mission: routeParams.mission-1, challenge: routeParams.challenge-1});
    } else {
      navigateToChallenge({level: 0, mission: 0, challenge: 0});
    }
  }
```
Does that make sense?

ekosmin: Wow yes, that is so much nicer! I had no idea you could have multiple route patterns like that. This is great, thanks Sam

ekosmin:
Per @kswenson's suggestion, the authoring document is now much more self explanatory- by converting arrays to objects, we are able to add `name` fields. The hardest part of all of these changes is updating the authoring documents specified in all of the test cases...

This change opens the door to add more application-specific fields down the line, if desired. There are more possible changes (e.g. moving the authoring utils into a module), but for now I'm going to try and wrap up this PR.

kswenson:
Clarification: the change isn't so much converting arrays to objects as it is embedding arrays as properties within objects (rather than relying on nested arrays). The `levels`, `missions`, and `challenges` arrays still exist, but their meaning is clearer when they are labeled as such within objects.

src/code/containers/challenge-container-selector.js
kswenson: Looking at this, another enhancement that comes to mind would be allowing partial route specification, e.g. `#/level/mission` would go to the first challenge of the specified mission. The relevance to this code is that to support `#/level` as a shortcut for the first mission of the specified level, then the corresponding line in `app.js` would be something like:
```
<Route path="/(:challengeIdOrLevel)" component={ChallengeContainerSelector} />
```
and then the `ChallengeContainerSelector` could test the type of the value to determine whether it's a challengeId (string) or level (number). That's out of scope for this PR, however, so I'll write it up as a PT story.

src/code/utilities/authoring-utils.js
```diff
@@ -0,0 +1,51 @@
...
+          challenges = authoring.challenges;
+    return challengeId && challenges ? authoring.challenges[challengeId] : null;
+  }
+
+  static getChallengeId(authoring, routeSpec) {
+    const levels = authoring.application.levels,
+          missions = levels ? levels[routeSpec.level].missions : null,
+          challenges = missions ? missions[routeSpec.mission].challenges : null,
+          challenge = challenges ? challenges[routeSpec.challenge] : null;
+    return challenge ? challenge.id : null;
```
ekosmin: This is a good point- I hadn't actually realized `return challenge && challenge.id;` would return null anyway, so I agree it's the cleaner solution here.

kswenson: If you just replaced the last one it would return null anyway. If you replaced all of the ternaries, it could potentially return undefined if levels were undefined initially.

PR #97, opened 2017-03-24
Fix gulp.watch issues

kswenson:
- New/deleted files now trigger rebuilds (add `{ cwd: './' }` argument) [#142103931]
- ~~`watch-fast` task now watches `resources` directory~~

Having `watch-fast` watch the `resources` directory turned out to be problematic because the resources task copies the entire resources folder which is slow and triggers excessive numbers of reloads, so it has been removed from this PR.

### DIFF
--- a/gulp/config.js
+++ b/gulp/config.js
@@ -1,11 +1,11 @@
-var src = './src',
-    bower = './bower_components',
-    node = './node_modules',
-    examples = './examples',
-    publicRoot  = './public',
-    publicExamples  = './public/examples',
-    publicGeniverse = './public/gv2',
-    dist = './dist';
+var src = 'src',
+    bower = 'bower_components',
+    node = 'node_modules',
+    examples = 'examples',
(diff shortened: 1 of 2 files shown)


ekosmin:
After having this in your environment for a while, did you decide if this PR is worth merging, @kswenson?

kswenson:
Good question. Under certain circumstances when there are a number of changes, such as when changing branches, the resources watcher seems to trigger two or three rebuilds for a bunch of resource files, even files that haven't changed at all. I think the ideal solution is some form of debouncing, so that only one rebuild is triggered regardless of the number of changes that occur. There is some discussion of debouncing approaches, but I haven't had time to pursue any of them. Therefore, my inclination has been to _not_ merge this for the moment. Another approach, which might be quicker than solving the debouncing issue would be to disable the resource file watcher but enable the new/deleted file detection for JS/CSS files. But watching resources properly would help to, so it'd be nice to figure out the debouncing and be done with it.

PR #98, opened 2017-03-24
Fix route navigation bug in PR#94

kswenson:
- use new `routeParams` in `componentWillReceiveProps` rather than `this.props.routeParams`
- add missing propType declaration for `challengeId` (in the neighborhood)


ekosmin:
Ah I see I confused the props, thanks for this fix Kirk

PR #100, opened 2017-03-27
Creating general animated sprite and animating hatching egg

ekosmin:
As opposed to the animation test PR [here](https://github.com/concord-consortium/geniblocks/pull/99), this PR is intended to merge into master. Here I implement and use a generic sprite component to animate a hatching egg.

Note that this PR is incomplete, because the sprite sheet is too large to upload to GitHub. I'm waiting on an updated asset from FableVision before these changes are usable.


sfentress:
> Note that this PR is incomplete, because the sprite sheet is too large to upload to GitHub.

?? Wow, how big is it?

src/code/components/animated-sprite.js
```diff
@@ -0,0 +1,42 @@
+import React, { PropTypes } from 'react';
+const tweenState = require('react-tween-state');
```
kswenson: You need to do a `yarn add react-tween-state` and commit the `package.json` and `yarn.lock` files as well.

src/stylus/challenges.styl
```diff
@@ -445,11 +445,9 @@ body
...
 
   .egg-image
-    width: 103px
-    padding: 34px
-    &.matching
-      margin-top: 0px
-      width: 70px
+    margin-top: 75px
+    margin-left: -10px
+    transform: scale(.7)
```
kswenson: I believe these changes will affect the old layout as well as the new.

src/code/components/animated-sprite.js
ekosmin: Thank you for this Kirk! I was wondering how to add the package. Also moved the CSS changes into an `fv-layout-c` section so now all layouts still look good.

PR #102, opened 2017-03-30
Adding generic counter, and counting score where necessary

ekosmin:
Since the HUD will have a number of counters (e.g. move, trial and score), I pulled out the counter into a generic component. I also added information to the `egg-sort-template` that allows it to show a maximum score, and also conveniently hide the score on challenges where it's not used.


src/code/fv-components/counter.js
```diff
@@ -0,0 +1,26 @@
+import React, {PropTypes} from 'react';
+import classNames from 'classnames';
+import t from '../utilities/translate';
+
+const CountView = ({countTitleText, className, currCount, maxCount}) => {
+
+  const countLabel = t(countTitleText),
```
ekosmin: I'm worried clients might not pass in a translatable string here...is there a better way to encourage that?

src/code/templates/fv-egg-sort-game.js
```diff
@@ -230,9 +230,10 @@ export default class FVEggSortGame extends Component {
     animation: null,
     animatedComponents: [],
     eggs: []
-  }
+  };
 
-  static backgroundClasses = 'fv-layout fv-layout-d'
+  static backgroundClasses = 'fv-layout fv-layout-d';
+  static maxScore = 8;
```
ekosmin: Of course this should be based on the number of drakes, but since all spec'd challenges have 8 eggs I figure this is fine for now. I realize I could initialize the variable below in `componentWillMount` based on the size of `drakes`- is that cleaner?

kswenson: I think it would be better to have this be state that is initialized by the template rather than static data. Given the fact that the egg sort game will also need to count `incorrect` responses, it might make sense to have a `score` object in the state that has properties for `correct`, `incorrect`, and 'maxScore`. Then if other templates require additional scoring options, there'd be a logical place to put them.

PR #104, opened 2017-03-31
Updating fv-chromosome component

ekosmin:
New art is coming in for the chromosomes, so I started to add functionality for dynamically placed labels.


src/stylus/fablevision/fv-gene-label.styl
```diff
@@ -0,0 +1,40 @@
...
+  .geniblocks.fv-gene-label
+    font-family: 'Yanone Kaffeesatz'
+    font-size: 1.5rem
+    color: white
+    position: absolute
+    display: flex
+    flex-direction: row
+    justify-content: flex-end
+
+    &.m
```
ekosmin: This is the biggest problem: at the moment, I'm placing the labels by eye, rather than the gene locations given in the chromosome. I chose this route because the assets I currently have are not based on the BioLogica definitions. It might be worth discussing this a bit more

kswenson: If the stripes are placed on the chromosomes for artistic reasons (e.g. to space them out in an artistically pleasing manner), then the labels should be placed relative to the art rather than according to the BioLogica species definition. If the stripes are meant to be biologically accurate, then they and their labels should be placed dynamically according to the BioLogica species definition. My inclination would be to do the latter for flexibility as well as accuracy, but ultimately that's a pedagogical determination rather than an engineering one. In the former case, with artistically placed stripes, specifying the positions of the labels in CSS may well be appropriate, as long as the JavaScript to set the appropriate classes is straightforward.

ekosmin:
I'm going to close this for now until we figure out what's happening with the chromosome stripes

PR #111, opened 2017-04-07
Enable dragging the user drake to the stable in the FVEggGame template

kswenson:
[PR#109](https://github.com/concord-consortium/geniblocks/pull/109) implemented the basic infrastructure for draggable drakes and droppable stables, but didn't enable it anywhere since we may not actually want those features. This PR actually enables draggable drakes and a droppable stable in the `FVEggGame` template, but will not be merged since those features aren't necessarily desirable. The techniques used here may be useful for reference, however, so this PR will be closed without merging but the code will be available for future reference.


kswenson:
Closing without merging since the code is for demonstration only.

PR #112, opened 2017-04-07
Implement most of FableVision Layout A

kswenson:
- Add FVGenomeChallenge template with layout A background assets from FableVision
- Add Your Drake label; remove instructions
- Add user drake surround image
- Add Hatch Drake button
- Add target drake gizmo
- Support variants with visible and hidden user drakes
- Add egg hatch animation
- Add target drake gizmo animation
- Increase size of genome label popup text (temporary until editable FableVision chromosomes are implemented)
- Add sex change toggle button
- Make strings localizable


src/code/templates/fv-genome-challenge.js
```diff
@@ -10,6 +10,23 @@ import { generateTrialDrakes } from '../utilities/trial-generator';
 const userDrakeIndex   = 0,
       targetDrakeIndex = 1;
 
+class YourDrakeView extends React.Component {
```
ekosmin: I like these internal components

src/code/templates/fv-genome-challenge.js
```diff
@@ -49,13 +54,36 @@ class YourDrakeView extends React.Component {
   }
 }
 
+/*
+ * TargetDrakeView
```
ekosmin: Not sure these comments are necessary

src/code/fv-components/fv-change-sex-toggle.js
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Stateless functional React component for displaying male/female change buttons
+ * The appearance of the buttons is currently entirely controlled via external CSS.
+ * @param {integer} sex - [BioLogica.MALE, BioLogica.FEMALE] currently selected button
+ * @param {function} onChange(evt, sex) - callback to be called when use clicks to change sex
+ */
+const FVChangeSexToggle = ({sex, style={}, onChange}) => {
+  const isMaleSelected = sex === BioLogica.MALE,
+        isFemaleSelected = sex === BioLogica.FEMALE,
```
kswenson: The `classNames()` calls below just seemed clearer (and less male-centric) with two variables rather than one negated variable.

src/code/components/animated-sprite.js
```diff
@@ -22,17 +23,31 @@ const FVAnimatedSprite = React.createClass({
...
+    this.setState({ started: true });
+  },
+
+  componentDidMount() {
+    if (!this.props.waitForTrigger)
+      this.animate();
+  },
+
+  componentWillReceiveProps(nextProps) {
+    if (!this.state.started && !nextProps.waitForTrigger)
```
kswenson: The current code guarantees that the animation can only be triggered once, which was the behavior I was aiming for. Yours allows the client to trigger multiple animations by repeatedly toggling `waitForTrigger`. After thinking about it a bit, your change could lead to a means to support looped sprite animations, so I'll make the change.

src/code/fv-components/egg-hatch-drake.js
```diff
@@ -0,0 +1,53 @@
...
+    drakeClasses: PropTypes.string,
+    drake: PropTypes.object,
+    hatchStarted: PropTypes.bool,
+    hatchComplete: PropTypes.bool,
+    onHatchComplete: PropTypes.func
+  }
+
+  constructor(props) {
+    super(props);
+    this.state = {
```
kswenson: Interesting to think about, but I'm inclined to leave this as is. The two props/state variables are what the only client is already maintaining and the client code is simple. Requiring that the client code re-instantiate the component and/or update a `key` seems like pushing responsibility to the client that is better handled by the component in this case. The component can only track its own `hatchComplete` state when it's actually performing the animation. The `hatchComplete` prop allows the client to skip the animation and have the component state be maintained appropriately.

src/code/templates/fv-genome-challenge.js
```diff
@@ -67,15 +68,35 @@ class YourDrakeView extends React.Component {
...
+          completeView = <div className='target-drake-gizmo' style={completeDisplayStyle}>
+                            <div id="target-drake-label">Target Drake</div>
+                            <OrganismView id="target-drake" org={org} width={300} />
+                          </div>;
     return (
-      <div className='target-drake-gizmo'>
-        <div id="target-drake-label">Target Drake</div>
-        <OrganismView id="target-drake" org={org} width={300} />
+      <div className='target-drake-view'>
+        { showIntro && !introComplete ? introAnimation : null }
```
kswenson: It may not currently ever be false, but I was thinking we might want to only show it the first time a the layout was visited, for instance, in which providing some control over its appearance would make sense.

src/code/fv-components/fv-change-sex-toggle.js
ekosmin: Makes sense! I'm all for less male-centric variables, and since they're `const` I guess there's no danger of having them get out of sync

src/code/templates/fv-genome-challenge.js
kswenson: Certainly not necessary, but when there are multiple components in a file, I think it helps to have visual breaks for scrolling/navigation. They also provide a place for comments about the role of the subcomponents, although I'm not using them for that purpose in this case.

src/code/fv-components/egg-hatch-drake.js
ekosmin: That makes sense about simplifying the client code, I can see using one component per challenge in that case.

I think I realized what's really making me uncomfortable about the code is that `hatchComplete` actually means multiple things. The `hatchComplete` prop is really used to communicate whether or not to skip the animation, and the `hatchComplete` state is used to track if the animation is finished. I would advocate for changing the name of the prop to something like `skipAnimation`, and leaving the rest of the code the way it is.

This clarifies the use of each variable in my mind, and lines them up better with the way they're used. What do you think?

kswenson: `hatchComplete` indicates whether the drake is hatched. That state can be achieved either by showing the animation and detecting when it has completed, or by the client stating that the drake is already hatched, in which case the animation is not required. I'm still inclined to leave things as they are.

ekosmin: The way I see it, the `hatchComplete` prop does not indicate whether a drake is hatched. In the case where the level hides the user drake, `hatchComplete` is *always* passed as false, even though the drake is sometimes hatched. This feels like a contradiction to me.

If it only seems confusing to me, then I support your decision to leave it as is!

src/code/templates/fv-genome-challenge.js
```diff
@@ -0,0 +1,218 @@
...
+export default class FVGenomeChallenge extends React.Component {
+
+  static backgroundClasses = 'fv-layout fv-layout-a'
+
+  state = {
+    hatchStarted: false
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (!this.props.showUserDrake && !this.props.userDrakeHidden && nextProps.userDrakeHidden)
```
ekosmin: I realized this is another confusing piece- `userDrakeHidden` and `showUserDrake` sound like they should be opposites. In reality, `showUserDrake` is referring to whether we're on a level that shows the drake, where `userDrakeHidden` refers to the current state of a level where we don't show the user drake by default.

kswenson: This naming goes back a lot further than this PR, so I'm not inclined to change it at this time.

src/code/fv-components/egg-hatch-drake.js
kswenson: That argument is compelling to me. I've pushed such a commit.

sfentress: Randomly playing with generating errors to test Rollbar, and I came across an error here and found this thread.

If a user is at #/6/1/5 and navigates to #/6/1/6, an error is thrown, because the egg hatching animation starts on the navigation event, finishes, and then errors because #/6/1/6 does not contain a target drake to check against.

Obviously, the egg hatching animation should not happen on the navigation event, but it does because #6/1/6 shows the drake, so `hatchStarted` gets set to true in this same view, and in the `componentWillReceiveProps` function nothing notices that `skipHatchAnimation` is also true. So the animation starts.

The simple way to fix this is to look at `skipHatchAnimation`:
```
componentWillReceiveProps(nextProps) {
    if (this.state.hatchStarted !== nextProps.hatchStarted)
      this.setState({ hatchStarted: nextProps.hatchStarted, hatchComplete: nextProps.skipHatchAnimation || false });
  }
```
However, without having looked too closely at this code or the conversations around it, I wonder if it wouldn't just be easier to have a `showDrake` or something property, that can skip all this logic. It seems strange to have to set two properties, `hatchStarted=true, skipHatchAnimation=true` when what you really mean is "just show the drake and pretend the egg doesn't exist," or `showDrake=true`.

PR #113, opened 2017-04-11
Distinct chromosome images

ekosmin:
In this change, chromosomes 1, 2, X and Y all now use images with different patterns of stripes on them. Before, we were passing booleans to tell if a chromosome was X or Y, but I've decided to pass around chromosome IDs instead. I also moved all the logic that turns those IDs into classes into `fv-chromosome-image` where it feels like it belongs. 

I included some other small changes here- updating the chromosome images in the gametes and fixing the target drake z-indexing issue.


src/stylus/fablevision/fv-chromosome.styl
```diff
@@ -15,24 +15,30 @@
     height: 122px
     width: 20px
 
+    &.one
+      background: url('../resources/fablevision/layout-c/chromosome_1.png')
```
kswenson: This occurs in many places, but I think `background-image` would be preferred here and elsewhere ([CSS Shorthand Syntax Considered an Anti-Pattern](https://csswizardry.com/2016/12/css-shorthand-syntax-considered-an-anti-pattern/)).

src/code/components/chromosome.js
```diff
@@ -90,7 +82,7 @@ const ChromosomeView = ({chromosome, org, ChromosomeImageClass=ChromosomeImageVi
     <div className="geniblocks chromosome-container" onClick={ handleSelect } >
       <div className={ containerClass }>
         <div className="chromosome-allele-container" id={chromId} style={displayStyle}>
```
ekosmin: I see what you mean- looking again, I see I left out the `orgName` in the `id`. I agree with your point above, that IDs have a lot of information shoved into them and it's not clear how it's all used. For now, I'll pull what I can into utility functions as you suggest so it's easier to address in the future, which feels like a step in the right direction.

ekosmin:
Ready for another round of review, this commit is much nicer! Your comments made me realize we had a bug before where empty child chromosomes actually did have duplicate `id`s. This is no longer a problem: `FVChromosomeImageView` now only requires a `name` to render, which does not need to be unique, and which is returned by a utility function. `ChromosomeView` uses the `name` of a chromosome to compute a unique `id` using the `orgName`.  

One side effect of these changes is that the old `egg-game` does not work, since I've pulled X/Y chromosome logic out of `ChromosomeView`. I'm tempted to bother leaving those bugs rather than updating `ChromosomeImageView`, let me know what you think.

ekosmin:
All changes requested here have been made, including the `background` -> `background-image` changes. Rather than merging, I'm going to wait on this PR to pass, which contains this PR in it:

https://github.com/concord-consortium/geniblocks/pull/114

PR #114, opened 2017-04-20
BioLogica-based Chromosome Labeling

ekosmin:
Here is a mega-PR that builds on the work done here (I rebased off that branch). All comments from the previous PR have been fixed here: 

https://github.com/concord-consortium/geniblocks/pull/113

The big changes here are that chromosome stripes are now drawn dynamically, based on their BioLogica definitions. To do this, we pass `chromosomeDescriptors` into the `FVChromosomeImageView`s which draw stripes from BioLogica. Those descriptors contain a chromosome name and side- information which gets stored in each rendered chromosome on the page inside its `id`.

Also worth noting is that labels are manually offset here, and placed in a constant static position. This means that here, the forelimbs label is placed at a diagonal, and always will be.


src/code/templates/fv-egg-game.js
```diff
@@ -1095,7 +1094,7 @@ export default class FVEggGame extends Component {
                                   selectedChromosomes: fatherSelectedChromosomes };
       return <GenomeView className={parentGenomeClass}  {...uniqueProps} ChromosomeImageClass={FVChromosomeImageView}
                         small={ true } editable={false} userChangeableGenes={ userChangeableGenes } visibleGenes={ visibleGenes }
-                        onAlleleChange={ handleAlleleChange }
+                        onAlleleChange={ handleAlleleChange } isSelectedEmpty={true} chromosomeHeight={122}
```
ekosmin: This is the biggest hack of the PR :( In the original version, a selected chromosome was drawn in bold in the parent genome, and in the offspring genome. In the FV version, a selected chromosome is drawn empty in the parent, and filled in the offspring. I hacked in a flag in the parent genome that basically says, treat me as empty if I'm selected.

To fix this correctly, I'm assuming I should make some changes to the way the state gets updated, to correctly reflect the state of the genomes.

test/components/chromosome.js
```diff
@@ -5,13 +5,13 @@ describe("<ChromosomeView />", function(){
         chromosome = drake.getGenotype().chromosomes[1]["a"];
 
   it("should create a <div> tag with appropriate classes", function() {
-    const wrapper = shallow(<ChromosomeView org={drake} chromosomeName='1' side='a'/>);
+    const wrapper = shallow(<ChromosomeView chromosome={chromosome}/>);
```
ekosmin: This test was the only place I could find `ChromosomeView` taking in a `side` and `chromosomeName` instead of a `chromosome`. Since it always receives a `chromosome`, and this information is redundant in that case, I removed the alternate "constructor" for `ChromosomeView`

kswenson: Seems like a worthwhile simplification.

src/code/templates/fv-egg-game.js
kswenson: In the particular case of `FVEggGame`, the clients are `parentGenomeView()` which wants to show _unselected_ chromosomes as filled, while `parentHalfGenomeView()` wants to show _selected_ chromosomes as filled. So for example, `parentGenomeView()` would pass `selectedChromosomes: invertChromosomes(motherSelectedChromosomes)`. There's an argument for naming the prop `filledChromosomes` or something like that, rather than `selectedChromosomes` while in the area.

src/code/fv-components/fv-gene-label.js
```diff
@@ -0,0 +1,52 @@
...
+    const style = {marginTop: percentHeight * chromosomeHeight - stripeHeight * 3 + "px"},
+          stripeStyle = {height: stripeHeight + "px"};
+
+    let labelStyle = {marginTop: -stripeHeight * .8},
+        lineStyle = Object.assign({}, labelStyle);
+    
+    if (normalizedAllele === "fl") {
+      labelStyle.marginTop -= stripeHeight * 5;
+      lineStyle.marginTop -= stripeHeight;
+    }
```
kswenson: We talked this afternoon about the interplay between JS and CSS where line/label-drawing is concerned.  I'm concerned that this will lead to brittle code with higher-than-optimal coupling between the JS/CSS. Unfortunately, I don't have a clear idea of what a better solution would look like so I'm inclined to let this go for now with the proviso that as you work with the code moving forward and as you add support for other alleles that require special-casing, that you consider whether there's a way to consolidate more (all?) of the responsibility on one side of the fence or the other.

ekosmin:
First commit is up to fix the kludge- I'll get up another commit to address the other issues. My fix is similar to what we discussed- the client can now remove chromosomes from the mapping and they'll appear as empty

src/code/components/chromosome.js
```diff
@@ -83,10 +77,11 @@ const ChromosomeView = ({chromosome, ChromosomeImageClass=ChromosomeImageView, u
     <div className="geniblocks chromosome-container" onClick={ handleSelect } >
       <div className={ containerClass }>
         <div className="chromosome-allele-container" id={chromId} style={displayStyle}>
-          <ChromosomeImageClass isSelectedEmpty={isSelectedEmpty} small={small} empty={empty} bold={selected} chromosomeDescriptor={getChromosomeDescriptor(chromosome)}/>
-          { allelesContainer }
+          <ChromosomeImageClass small={small} empty={empty} bold={selected} chromosomeDescriptor={chromosomeDescriptor}/>
+        </div>
+        <div className="labels">
```
ekosmin: I'm always leaving the labels div in here, although for empty chromosomes the div will be empty. This helps with formatting, since now chromosomes have the same length even as they gain or lose labels

src/code/components/chromosome.js
```diff
@@ -2,21 +2,41 @@ import React, {PropTypes} from 'react';
...
+ * @param {string} orgName - the unique name of the organism which the chromosome is a part of (e.g. "targetMother")
+ * @param {number} height - the height in pixels of the rendered chromosome
+ * @param {object} displayStyle - this style will be applied to the actual chromosome, as opposed to its labels
  */
-
-const ChromosomeView = ({chromosome, ChromosomeImageClass=ChromosomeImageView, userChangeableGenes = [], visibleGenes = [], hiddenAlleles = [], small = false, editable = true, selected = false, onAlleleChange, onChromosomeSelected, showLabels = true, showAlleles = false, labelsOnRight = true, orgName, isSelectedEmpty, height, displayStyle = {}}) => {
+const ChromosomeView = ({chromosome, chromosomeDescriptor, ChromosomeImageClass=ChromosomeImageView, userChangeableGenes = [], visibleGenes = [], hiddenAlleles = [], small = false, editable = true, selected = false, onAlleleChange, onChromosomeSelected, showLabels = true, labelsOnRight = true, orgName, height, displayStyle = {}}) => {
+  chromosomeDescriptor = chromosomeDescriptor || getChromosomeDescriptor(chromosome);
```
ekosmin: Adding a `chromosomeDescriptor` here is necessary since empty chromosomes are now passed as null. The descriptor lets the `ChromosomeImage` render the correct empty chromosome

test/components/allele.js
```diff
@@ -1,28 +0,0 @@
-import AlleleView from '../../src/code/components/allele';
```
ekosmin: I couldn't find uses of this anywhere in the codebase, so I removed it and the tests that referred to it

src/code/fv-components/fv-gene-label.js
ekosmin: 👍 yes this sounds good- I'm going to try and keep it all in CSS in the future

src/code/fv-components/fv-gene-label.js
```diff
@@ -0,0 +1,52 @@
...
+    var chromStart;
+    Object.keys(BioLogica.Species.Drake.geneList).forEach((geneName) => {
+      const gene = BioLogica.Species.Drake.geneList[geneName];
+      if (gene.alleles.indexOf(allele) > -1) {
+        chromStart = gene.start;
+      }
+    });
+    let bioChromHeight = BioLogica.Species.Drake.chromosomesLength[chromosomeDescriptor.name],
+        stripeHeight = chromosomeHeight * .03,
+        percentHeight = chromStart/bioChromHeight;
```
ekosmin: Since this deals so much with BioLogica code pulling it out is a good idea. I ended up keeping the `stripeHeight` calculation in `fv-gene-label`, since it's not actually using the length of the gene from BioLogica (which would be smaller than desired)

src/code/fv-components/fv-chromosome-image.js
```diff
@@ -1,29 +1,60 @@
...
+}
+
+const FVChromosomeImageView = ({small=false, bold=false, empty=false, isSelectedEmpty, chromosomeDescriptor, animationStyling}) => {
+  function chromosomeDescriptorToClass(chromosomeDescriptor) {
+    if (!chromosomeDescriptor) {
+      return null;
+    } else if (chromosomeDescriptor.side === 'y') {
+      return 'chromosome-y';
+    } else if (chromosomeDescriptor.side === 'x') {
+      return 'chromosome-x';
```
ekosmin: Yes, good catch, all set!

src/code/templates/fv-egg-game.js
```diff
@@ -991,6 +992,28 @@ export default class FVEggGame extends Component {
       onResetGametes();
     };
 
+    function getUnselectedChromosomesMap(chromosomes, selectedChromosomes) {
```
ekosmin: Here's where I'm now building up a map of unselected chromosomes. In general, the code is much cleaner now, even though I think it might have been cleaner to null a full mapping instead of building it from nothing here

PR #116, opened 2017-04-20
Adding custom dropdowns

ekosmin:
Adding custom dropdown component to `fv-gene-label`. Also manually tweaking styling for labels of the horn gene.

Note that this is branching off of the `updateChromosomes` branch, so I would recommend reviewing that PR first: 

https://github.com/concord-consortium/geniblocks/pull/114


src/stylus/gene-label.styl
```diff
@@ -1,7 +0,0 @@
-.geniblocks.allele
```
ekosmin: I couldn't figure out where this styling is used, and it was throwing off my drop downs, so I removed it. Do you know where I could see it used to maybe restrict this styling?

src/stylus/react-selectize.styl
```diff
@@ -0,0 +1,929 @@
+.example .react-selectize.multi-select .removeable-emoji {
```
ekosmin: Apparently, I should be able to add this line to add all styling for `react-selectize`:

@import 'node_modules/react-selectize/themes/index.css'

However, this wasn't working for me, so I copied and pasted all the `css` from the `react-selectize` examples page...

src/stylus/gene-label.styl
kswenson: In `GeneLabelView`:
```
    return (
      <div className="geniblocks gene-label allele editable">
        <select value={ allele } onChange={ onAlleleChange }>
          { alleleOptions }
        </select>
      </div>
    );
```

src/stylus/react-selectize.styl
kswenson: You should be able to tell Browserify about it so that it gets bundled with the rest of the CSS.

ekosmin: Cool! I've gotten that bundled now, and the dropdowns work without this ugly file

PR #117, opened 2017-04-20
ITS Updates

ekosmin:
A handful of different ITS updates:
* Adding information to `submitEgg` action for the egg sorting game
* Adding integration with the Portal. If launched from an activity there, the student and classroom IDs are now pulled out of the URL.
* ITS now hints after a student has bred two drakes without completing a challenge. Functionally, this means that after a single reset, the 'reset' and 'submit' buttons to the same thing

Turns out implementing the last bullet point was simpler than expected- I gave `resetGametes` a flag to optionally increment the number of moves, and check `moves` when the 'reset' button is hit.

### DIFF
--- a/src/code/actions.js
+++ b/src/code/actions.js
@@ -268,16 +268,20 @@ function getEarnedCoinString(state) {
   return ["~ALERT.NEW_PIECE_OF_COIN", scoreString];
 }
 
+function getVisibleGenes(state, routeSpec) {
+  const visibleGenes = AuthoringUtils.getChallengeDefinition(state.authoring, routeSpec).visibleGenes;
+  return visibleGenes && visibleGenes.split(", ");
+}
+
 function _submitDrake(targetDrakeIndex, userDrakeIndex, correct, state) {
   const incrementMoves = !correct,
         targetDrakeOrg = GeneticsUtils.convertDrakeToOrg(state.drakes[targetDrakeIndex]),
         userDrakeOrg = GeneticsUtils.convertDrakeToOrg(state.drakes[userDrakeIndex]),
         initialDrakeOrg = state.initialDrakes[userDrakeIndex] ? GeneticsUtils.convertDrakeToOrg(state.initialDrakes[userDrakeIndex]) : null,
         routeSpec = state.routeSpec,
-        visibleGenes = AuthoringUtils.getChallengeDefinition(state.authoring, routeSpec).visibleGenes,
         // TODO: figure out whether ITS really wants "editableGenes" or "visibleGenes"
         // because it doesn't make sense to log the latter as the former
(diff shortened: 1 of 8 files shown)


src/code/actions.js
```diff
@@ -268,16 +268,20 @@ function getEarnedCoinString(state) {
   return ["~ALERT.NEW_PIECE_OF_COIN", scoreString];
 }
 
+function getVisibleGenes(state, routeSpec) {
+  const visibleGenes = AuthoringUtils.getChallengeDefinition(state.authoring, routeSpec).visibleGenes;
```
sfentress: Should the `userChangeableGenes` be added to this list? For all the trials involving the user changing pulldowns, this will result in sending `editableGenes: null` in the `submitDrake` action (or if the authors add an unchangable gene label, it will send just that label). I feel like this can't be what is wanted.

In the basket challenges, clearly we just want the visible genes. So I could see two options: (1) merge the two arrays, to submit all genes the user can see, or (2) send the userChangeableGenes if there are any, and the visibleGenes if there aren't.

sfentress: ...possibly the real error was my calling the list "visibleGenes" instead of "staticGenes" or something.

kswenson: Eli and I have talked about this. I even wrote a PT story for talking with Rob about what ITS actually wants/needs and cleaning up the code. I'd be game for changing `visibleGenes` to `staticGenes` or something else if it makes things clearer, but I think what Rob says the ITS actually wants should be factored into any naming discussion. The name(s) should make sense from an authoring standpoint as well as a logging standpoint.

kswenson:
>I feel like we keep seeing more and more stuff in the actions that isn't needed by the reducers, in order to satisfy the needs of logging and ITS

Eli and I have talked about this as well. I think you're suggestion of putting additional information required only for logging into a separate subfield would be a step in the right direction. Going even further, I think there's an argument for saying that the templates/action creators shouldn't be responsible for providing the additional information, but rather the middleware should gather what it requires on its own. Since the ITS middleware has access to the state, the ITS/logging middleware are capable of pulling out any additional state that they require in a manner that is completely isolated to the ITS/logging middleware code. But that degree of architecture change should probably be a separate PT story for later refactoring. In the meantime, I'd be in favor of packaging it in a separate subfield as you suggest.

sfentress:
> Going even further, I think there's an argument for saying that the templates/action creators shouldn't be responsible for providing the additional information, but rather the middleware should gather what it requires on its own.

:+1: That would be the ideal solution, in the long run.

ekosmin:
Added PT stories for logging information in middleware instead of the templates, and for figuring out the `visibleGenes` vs `editableGenes` vs `staticGenes` confusions. Thanks all!

PR #118, opened 2017-04-24
Adding index page to navigate to all challenges within the same URL

ekosmin:
I put together a very simple navigation page to use for a classroom test, now confirmed to begin on May 8th. The idea is that a page like this is helpful because it will automatically pass along URL parameters from the Portal, unlike the current examples index page we have at a different URL. It should now be easy to create the Portal activity for the classroom test.

### DIFF
--- a/src/code/containers/challenge-container-selector.js
+++ b/src/code/containers/challenge-container-selector.js
@@ -2,7 +2,7 @@ import React, { Component, PropTypes } from 'react';
 import { DragDropContext } from 'react-dnd';
 import HTML5Backend from 'react-dnd-html5-backend';
 import { connect } from 'react-redux';
-import { navigateToCurrentRoute } from '../actions';
+import { navigateToCurrentRoute, navigateToChallenge } from '../actions';
 import ChallengeContainer from './challenge-container';
 import FVChallengeContainer from './fv-challenge-container';
 import AuthoringUtils from '../utilities/authoring-utils';
@@ -12,8 +12,12 @@ function hasChangedRouteParams(currentRouteSpec, routeParams) {
         currLevelStr = String(currLevel + 1),
         currMissionStr = String(currMission + 1),
         currChallengeStr = String(currChallenge + 1);
-  return (routeParams.mission && routeParams.challenge && routeParams.level &&
+  return (isValidRouteParams(routeParams) &&
         ((routeParams.mission !== currMissionStr) || (routeParams.challenge !== currChallengeStr) || (routeParams.level !== currLevelStr)));
+}
+
+function isValidRouteParams(routeParams) {
+  return routeParams.mission && routeParams.challenge && routeParams.level;
 }
 
 function mapContainerNameToContainer(containerName) {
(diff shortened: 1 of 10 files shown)


sfentress:
How hard would it be to move the index page to its own route, instead of making it a challenge template? Using a template for this feels like a bit of a hack, which is fine if this is very temporary, or being done in a rush, but it seems like this (or, rather, something very like it) will probably remain in the code base (i.e. as we add home screens, etc.).

ekosmin:
Thanks for these comments @sfentress, and @kswenson for taking the time to talk to me about this. I have a new version up which I hope addresses your concerns!

Sam, you're absolutely right that the original version was a hack- I displayed the page as a challenge because it was the only way I was familiar with. After taking a closer look, I figured out how to give the page its own route, and free it from any challenge container.

@kswenson raised the point that, in the past, we had created temporary navigation pages for different classroom tests, like this one: http://geniverse.concord.org/drakes-traits/. I built this index page into the app at @scytacki's suggestion, since it makes it simpler to pass URL parameters on from the portal. If we need to keep changing the navigation page for classroom tests, it might make sense to remove it from this project, like it was in the past. For now, though, I'm tempted to keep something like this as a base for future navigation pages, like Sam mentioned above.

ekosmin:
@kswenson or @sfentress, I'd appreciate it if you'd also be able to look at my most recent commit. I realized that after my change to routing, the buttons no longer worked as expected. They would all navigate to the correct level, but then immediately navigate to level 1.1.1. I believe that's because the navigation page moved to the correct level, then the challenge container selector renders and sees that the next route spec equals the current route spec, so it resets to 1.1.1.

I don't fully understand why that's happening. I expect the reset behavior is there to wrap levels back to the beginning, but since that behavior doesn't seem very important if we have a navigation page, I removed it so now there are no issues. Is there a cleaner solution I'm overlooking?

src/code/containers/challenge-container-selector.js
```diff
@@ -52,8 +56,10 @@ class ChallengeContainerSelector extends Component {
     if (routeParams.challengeId) {
       routeParams = AuthoringUtils.challengeIdToRouteParams(authoring, routeParams.challengeId);
     }
-    if (hasChangedRouteParams(this.props.currentRouteSpec, routeParams)) {
-      navigateToCurrentRoute({level: routeParams.level-1, mission: routeParams.mission-1, challenge: routeParams.challenge-1});
+    if (isValidRouteParams) {
```
ekosmin: Alright, here are the changes we discussed @kswenson! Now, if route params are missing, we still navigate to a default level, and if we've already navigated to a level, we do nothing. 

FWIW, the `currentRouteSpec` when the challenge loads from a URL is set to 0.0.0

kswenson: The defaulting path is inaccessible as written: `isValidRouteParams` is a function that must be called -- `isValidRouteParams()`. Beyond the defaulting issue, when I enter the navigate URL I end up with `routeParams` that has a `challengeId` but not a `level`, `mission`, or `challenge`, which then fails in a different way. I think the code needs some more work before it's worth reviewing further. Given that I'm going to be gone for the next couple of days, that means Sam will have to do further reviews.

ekosmin: Man, good catch on the validity checking...that's where unit tests help. I don't believe your second comment is correct, though: this code path is never run when you enter the navigate URL, because we route directly to the `Navigation` component.

src/code/containers/navigation.js
```diff
@@ -0,0 +1,81 @@
...
+        </div>
+      </div>
+    );
+  }
+
+  static propTypes = {
+    navigateToChallenge: PropTypes.func.isRequired
+  }
+
+  static authoredDrakesToDrakeArray = function() {
```
sfentress: This isn't needed, right? It made me worry that `loadStateFromAuthoring` was being called on this route, but it isn't as far as I can tell, unless there's another way to access this route that uses `action.NAVIGATED`.

ekosmin: Good catch Sam! This was from an earlier version when the page was loaded as a template, so I should get rid of it now

PR #119, opened 2017-04-25
Adding a stable/create unique version to the genome challenge

ekosmin:
This commit replicates the original "playground" level, using the `fv-genome-challenge` template. This is done using a `challenge-type` flag, much like `fv-egg-game` uses to differentiate target matching vs. unique drake creation challenges.

I also updated the stable image here, to the smaller one which will be used in the re-organized layout.


src/code/utilities/lang/en-us.js
```diff
@@ -9,6 +9,8 @@ export default {
   "~EGG_GAME_3.INSTR_ITEM1": "Scope an egg to see a baby's chromosomes.",
   "~EGG_GAME_3.INSTR_ITEM2": "Click the correct basket for the egg.",
 
+  "~GAMETE_CHALLENGE.BUTTON.SAVE_DRAKE": "Save Drake",
```
kswenson: Shouldn't that be `GENOME_CHALLENGE.BUTTON`?

ekosmin:
Good point about the new assets- I'm going to lump that together in a PR with the chromosome changes as well. Thanks Kirk!

PR #122, opened 2017-05-01
Rollbar Integration

ekosmin:
Hi @scytacki, would you mind taking a look at our Rollbar integration? I just added the `<script>` to the index page, not sure if there's any more configuration to do.


sfentress:
As far as I can tell this is integrated well, but I'm not an expert on Rollbar so I could be missing something.

PR #125, opened 2017-05-07
Gamete breeding egg game

ekosmin:
This is a first pass at the gamete breeding version of the egg game. It uses the new gamete images in the new locations, using a similar strategy as before: the `GametePen` now takes in a `GameteImageClass` to choose the proper gamete. It doesn't use the new background, and I didn't follow the exact layout from Basecamp, but it's a step in the right direction- I'm hoping it's sufficient for the classroom test this week.


sfentress:
I think these changes look good. :+1:

There are a couple issues with the original code, which aren't touched by this PR, but might be worth doing something about now:

1. The `GAMETE_SELECTED` action has an itsLog that includes `action: ITS_ACTIONS.SELECTED`, but `ITS_ACTIONS.SELECTED` doesn't exist. This should probably be `ITS_ACTIONS.CHANGED_SELECTION` -- I don't know whether there's a functional difference between "selected" and "changed selection"
2. The user can still select gametes after an offspring has been bred. This results visually in an offspring with chromosomes under it which do not match it. It's also just confusing to be able to do this action with no actual effect. I think perhaps the `GAMETE_SELECTED` action should do nothing in the reducer if there is already an offspring.

ekosmin:
Thanks for taking a look @kswenson and @sfentress! Those are good points- I'll log them both in PT so they don't get dropped, I want to make sure to merge this for tomorrow's classroom test

PR #126, opened 2017-05-09
Moving dialog boxes to center of screen

ekosmin:
A small fix if you wouldn't mind taking a look @kswenson! Trudi noticed that dialog boxes aren't centered on the screen, and may move off entirely on smaller screens. 

I realized the issue is that the dialog boxes have a static offset from the top which isn't scaled, because it's not in the challenge container. I removed the offsets, so now the default `50%` is used. The fix looks fine, but there's a comment in the code that made me want a second pair of eyes:

```
// we use some pseudo random coords so nested modals
// don't sit right on top of each other.
````

I don't want these modals to get covered up, but in this case I'm fairly sure that won't happen.

### DIFF
--- a/src/code/actions.js
+++ b/src/code/actions.js
@@ -389,7 +389,6 @@ export function submitDrake(targetDrakeIndex, userDrakeIndex, correct, incorrect
             label: "~BUTTON.NEXT_TRIAL",
             action: "advanceTrial"
           },
-          top: "600px"
         };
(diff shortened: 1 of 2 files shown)


ekosmin:
Heh ok thank you for the background! It does seem like this is fine, especially since, as you mention, these messages will likely end up in the HUD soon.

PR #127, opened 2017-05-18
Changing initial drake generation to make sure it does not match the target drake

ekosmin:
For early genome challenges, it was possible for the target drake to match the initial drake exactly. To fix this, I changed the `trialGenerator` for one of the levels, and tweaked the `all-combinations` generation code. Now, it tries up to 100 times to generate 2 drakes with different phenotypes.

I'll admit, the randomization schemes have never been totally clear to me- I'm not sure who worked on each one. Does this seem like a reasonable change? As for the other genome challenges, I believe they avoid duplicates by more specifically outlining each trial.

### DIFF
--- a/src/code/utilities/trial-generator.js
+++ b/src/code/utilities/trial-generator.js
@@ -11,10 +11,27 @@ export function generateTrialDrakes(trialDef, trial=0) {
 
     let initialDrakeAlleles = generateComboDrakeAlleles(trialDef.baseDrake, trialDef.initialDrakeCombos, trial),
         targetDrakeAlleles  = generateComboDrakeAlleles(trialDef.baseDrake, trialDef.targetDrakeCombos,  trial),
-        initialDrake = {alleles: initialDrakeAlleles, sex: 0},
-        targetDrake = {alleles: targetDrakeAlleles, sex: 0};
+        initialDrake = new BioLogica.Organism(BioLogica.Species.Drake, initialDrakeAlleles),
+        targetDrake = new BioLogica.Organism(BioLogica.Species.Drake, targetDrakeAlleles);
 
-    return [initialDrake, targetDrake];
+    for (let numTrials = 0; numTrials < 100; numTrials++) {
+      if (arePhenotypesEqual(initialDrake, targetDrake)) {
+        initialDrakeAlleles = generateComboDrakeAlleles(trialDef.baseDrake, trialDef.initialDrakeCombos, trial);
+        initialDrake = new BioLogica.Organism(BioLogica.Species.Drake, initialDrakeAlleles);
+      } else {
+        break;
+      }
+    }
+    if (arePhenotypesEqual(initialDrake, targetDrake)) {
+      console.log("Failed to produce different initial and target drakes");
+    }
+
+    let initialDrakeSex = Math.trunc(2 * Math.random()),
+        targetDrakeSex = Math.trunc(2 * Math.random()),
+        initialDrakeObj = {alleles: initialDrakeAlleles, sex: initialDrakeSex},
+        targetDrakeObj = {alleles: targetDrakeAlleles, sex: targetDrakeSex};
+
+    return [initialDrakeObj, targetDrakeObj];
   }
   if (trialDef.type === RANDOMIZE_ORDER) {
     if (trialDef.drakes) {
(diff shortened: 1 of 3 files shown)


src/code/utilities/trial-generator.js
```diff
@@ -34,6 +51,16 @@ export function generateTrialDrakes(trialDef, trial=0) {
   return [];
 }
 
+function arePhenotypesEqual(drake1, drake2) {
```
sfentress: There's an easier way to do this: `drake1.getImageName() === drake2.getImageName()`.

Image names are by definition the phenotype.

sfentress:
This looks fine to me. I was initially hesitant about generating the random drakes and comparing them, and thought we could do something in the `generateComboDrakeAlleles` function to ensure we pick different columns for two drakes, but then I realized that they key is to make two phenotypically-different drakes, and we can't do that by looking at the alleles alone.

I would add a test for this. Actually, I did add a test for it, because I wanted to confirm for myself that it would work. Check if the tests below look good to you, and are testing the right thing, then you could add them to the first `describe` block of test/utilities/trial-generator.js:

```
  it('should generate different drakes if possible', () => {
    const spec = GeneticsUtils.convertDashAllelesObjectToABAlleles({
                  "type": "all-combinations",
                  "baseDrake": "T-T, W-W, Fl-Fl, Hl-Hl, h-h, A1-A1, C-C, B-B, D-D, rh-rh, Bog-Bog",
                  "initialDrakeCombos": [
                    ["M-M",  "m-m"]
                  ],
                  "targetDrakeCombos": [
                    ["M-M",   "m-m"]
                  ]
                }, ["baseDrake", "initialDrakeCombos", "targetDrakeCombos"]);
    // stochastic test, so try several times
    for (let tries = 0; tries < 10; tries++) {
      let drakes = generateTrialDrakes(spec, 0);
      // just compare alleles for this test, but actual code ensures diff phenotypes
      assert.notEqual(drakes[0].alleles, drakes[1].alleles);
    }
  });

  it('should fail gracefully if impossible to generate different drakes', () => {
    const spec = GeneticsUtils.convertDashAllelesObjectToABAlleles({
                  "type": "all-combinations",
                  "baseDrake": "T-T, W-W, Fl-Fl, Hl-Hl, h-h, A1-A1, C-C, B-B, D-D, rh-rh, Bog-Bog",
                  "initialDrakeCombos": [
                    ["M-M"]
                  ],
                  "targetDrakeCombos": [
                    ["M-M"]
                  ]
                }, ["baseDrake", "initialDrakeCombos", "targetDrakeCombos"]);
    let drakes = generateTrialDrakes(spec, 0);
    expect(drakes.length).toBe(2);
    expect(GeneticsUtils.isValidAlleleString(drakes[0].alleles)).toBe(true);
    expect(GeneticsUtils.isValidAlleleString(drakes[1].alleles)).toBe(true);
  });
```

Two questions/thoughts:

1. The authoring change will make those trials different from what it was previously. Trudi definitely liked having randomized traits for the things that didn't matter, and using the trial-generator eliminates that. Is having the trial-generator for that challenge a separate request that's being bundled into this PR, or was it just to try and get different drakes?
2. If you want to make different drakes even with the original authoring style, you're right that authors could just pin the traits down further. But if there are situations where that doesn't make sense for them, I think it could potentially still be handled. Maybe the best place to do it would be in `processAuthoredDrakes`. Maybe the template could send back indexes of drakes that should never be the same (or, heck, we could just assume than no authored drakes should ever be the same in one trial), and `processAuthoredDrakes`, and in the mapping that starts with `let drakes = authoredDrakesArray.map(...`, we could just check that each new drake doesn't match any existing drake (using your randomize-and-exit method if you can't make different drakes).

ekosmin:
@sfentress thank you for those tests, they work great! They fail as expected if I just have `arePhenotypesEqual` return `false`, and pass otherwise. Your questions are good ones:

1) Using a trial-generator was not part of any request- as you suggested, it was just to get different drakes. I chose to use it because it was being used in the "hidden drake" levels, so I could just copy and paste it without thinking too much about the specifics for each generated drake.

2) I like the idea of having `processAuthoredDrakes` ensure no drakes are identical, to limit the amount of information that needs to be locked in for drake definitions. Let me reach out to Trudi on the PT story for this PR to see what her thoughts are on this.

I feel like there's still a larger conversation to be had about randomization expectations...I really appreciate your comments here to help me think through the best ways to do it!

PR #128, opened 2017-05-18
Egg Sorting ITS Upgrade

ekosmin:
@kswenson, I'm sending this over to you, since we had discussed the communication from GV2 to ITS a bit in person. Ultimately, we pushed the responsibility to the ECD matrices to use the constant string keys in BioLogica, so no large-scale changes were required here or in BioLogica, and all the matrices are up-to-date at this point. In the end, this PR is mostly re-organizing some information into the form that the ITS expects.

### DIFF
--- a/src/code/templates/fv-egg-sort-game.js
+++ b/src/code/templates/fv-egg-sort-game.js
@@ -332,7 +332,10 @@ export default class FVEggSortGame extends Component {
     isSubmittedEggCorrect = selectedEgg && isEggCompatibleWithBasket(selectedEgg, basket);
     if (selectedEgg) {
       animationEvents.moveEggToBasket.animate(selectedEgg, selectedEggIndex, basketIndex);
-      onSubmitEggForBasket(eggDrakeIndex, basketIndex, isSubmittedEggCorrect, isChallengeComplete);
+      setTimeout(function() {
+        // Delay the submit until the egg is finished hatching
+        onSubmitEggForBasket(eggDrakeIndex, basketIndex, isSubmittedEggCorrect, isChallengeComplete);
+      }, 4000);
     }
     this.setBasketSelection([basketIndex]);
   }
--- a/src/code/utilities/genetics-utils.js
+++ b/src/code/utilities/genetics-utils.js
@@ -472,12 +472,18 @@ export default class GeneticsUtils {
     return {geneStart, chromosomeHeight: species.chromosomesLength[chromosomeName]};
   }
 
-  static convertGenesToPhenotype(genes) {
+  /**
+   * Given a gene string, returns an object representing the phenotypes expressed by that gene string. For example,
+   * 'a:w,b:w' is converted to {'wings': 'No wings'}.
+   */
+  static convertGeneStringToPhenotype(genes) {
+    // First, create a drake from the given gene string. This drake has the correct phenotype, but we only want the
+    // part of the phenotype specifically referenced in the gene string
     const species = BioLogica.Species.Drake,
           geneDrake = new BioLogica.Organism(species, genes[0], BioLogica.FEMALE),
           phenotype = geneDrake.phenotype.allCharacteristics;
 
-    // First, get a list of all alleles in the given gene string
+    // To get the phenotype, we first go through all possible alleles, and pull out the ones used in the genome string
     let relevantAlleles = [];
     Object.keys(species.alleleLabelMap).forEach((alleleName) => {
       if (alleleName !== "" && (genes[0].indexOf("a:" + alleleName) > -1 || genes.indexOf("b:" + alleleName) > -1)) {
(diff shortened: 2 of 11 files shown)


src/code/templates/fv-egg-sort-game.js
```diff
@@ -332,7 +332,10 @@ export default class FVEggSortGame extends Component {
     isSubmittedEggCorrect = selectedEgg && isEggCompatibleWithBasket(selectedEgg, basket);
     if (selectedEgg) {
       animationEvents.moveEggToBasket.animate(selectedEgg, selectedEggIndex, basketIndex);
-      onSubmitEggForBasket(eggDrakeIndex, basketIndex, isSubmittedEggCorrect, isChallengeComplete);
+      setTimeout(function() {
+        // Delay the submit until the egg is finished hatching
+        onSubmitEggForBasket(eggDrakeIndex, basketIndex, isSubmittedEggCorrect, isChallengeComplete);
+      }, 4000);
```
ekosmin: This is one other meaningful change I made - previously, this delay was handled in several places in the action, but I refactored it here to accommodate different templates using different hatch times.

test/actions/submit-egg.js
```diff
@@ -16,15 +16,11 @@ const types = actions.actionTypes,
       },
       submittedAlleles = "a:T,b:T,a:M,b:M,a:W,b:w,a:h,b:h,a:C,b:C,a:B,b:B,a:Fl,b:Fl,a:hl,b:hl,a:A1,b:A1,a:D,a:Bog,a:rh",
       acceptedAllelesCorrect = ["a:W", "b:W"],
+      acceptedPheotypeCorrect = {"wings": "Wings"},
       acceptedAllelesIncorrect = ["a:w,b:w"],
+      acceptedPheotypeIncorrect = {"wings": "No wings"},
```
kswenson: 2x: "Pheotype" -> "Phenotype"

test/actions/submit-egg.js
```diff
@@ -60,13 +56,13 @@ describe('submitEggForBasket action', () => {
...
-        acceptedGenotypes: acceptedAllelesCorrect,
-        acceptedSexes: [BioLogica.FEMALE, BioLogica.MALE],
-        basketLabel: submittedBasketLabel,
-        visibleGenes: ["wings", "arms"],
+        challengeCriteria: {
+          alleles: submittedAlleles,
+          sex: 0
+        },
+        userSelections: {
+          phenotype: acceptedPheotypeCorrect
```
kswenson: "Pheotype" -> "Phenotype"

test/actions/submit-egg.js
```diff
@@ -115,13 +111,14 @@ describe('submitEggForBasket action', () => {
...
-        acceptedGenotypes: acceptedAllelesIncorrect,
-        acceptedSexes: [BioLogica.FEMALE, BioLogica.MALE],
-        basketLabel: submittedBasketLabel,
-        visibleGenes: ["wings", "arms"],
+        challengeCriteria: {
+          alleles: submittedAlleles,
+          sex: 0
+        },
+        userSelections: {
+          phenotype: acceptedPheotypeIncorrect,
```
kswenson: "Pheotype" -> "Phenotype"

src/code/utilities/genetics-utils.js
```diff
@@ -471,4 +471,59 @@ export default class GeneticsUtils {
     });
     return {geneStart, chromosomeHeight: species.chromosomesLength[chromosomeName]};
   }
+
+  /**
+   * Given a gene string, returns an object representing the phenotypes expressed by that gene string. For example,
+   * 'a:w,b:w' is converted to {'wings': 'No wings'}.
+   */
+  static convertGeneStringToPhenotype(genes) {
```
ekosmin: This code got a little away from me, but I haven't spent time optimizing it, because I think specs are likely to change. This is an attempt to turn our basket definitions (which come in the form of allele strings) into what the ITS expects (i.e. accepted phenotypes).

For each basket, I end up passing over a mapping of the phenotypes they expect e.g. `{'wings': 'No wings'}` which I derive from the allele strings of the baskets. However, we're already realizing this is an insufficient way to describe the baskets - for example, it cannot describe a basket that accepts either 3 or 5 armor plates. Rob hacked a solution to this on his side, but if baskets get any more complex, I think it will be best to specify whatever information the ITS needs in the authoring doc instead of trying to derive it from allele strings here.

Happy to discuss this issue more, it's proven difficult to solve well.

sfentress: Is there a reason we have to specify baskets using alleles, and not traits?

It seems like making it traits instead would make everything a lot simpler.

ekosmin: I'm not sure, I agree using traits instead would be simpler. I didn't want to change all the grading code, but that's probably the correct solution. 

Rather than saying something like:

`{ "label": "Drakes with forelimbs and hindlimbs", "alleles": ["Fl-, Hl-","Fl-, -Hl","-Fl, Hl-","-Fl, -Hl"] }`

It would probably be simpler to write:

`{ "label": "Drakes with forelimbs and hindlimbs", "phenotypes": [{"forelimbs": "Forelimbs", "hindlimbs": "Hindlimbs"] }`

ekosmin: I'll make @kswenson's changes for now, and leave a TODO to clean this up the next time we need to make ITS changes

PR #130, opened 2017-05-25
Color levels and randomization changes

ekosmin:
Hi @sfentress! I added two new levels which allow students to play with color in a Genome Challenge. Those changes were small, but the level spec required some more involved updates I'd appreciate your feedback on.

For the first new level, students have to match 5 target drakes over 5 trials - one each of 5 colors, in a random order. For the second level, students match 3 drakes, picking out of the 5 different colors at random. To accomplish this, I added a new `numTrials` field to the authoring doc and state, which allows you to specify any number of trials, and then choose a random subset of them (3 out of a possible 5, in this case). It actually simplified some pieces of code throughout the project.

Let me know what you think!


sfentress:
:+1: This looks great. It's nice how localized the change was able to be.

sfentress:
I hit the Merge button, and hadn't realized you had requested to merge into `mayDay1`. Just confirming here that that's what you meant.

PR #131, opened 2017-06-02
Color levels

ekosmin:
This was already reviewed by @sfentress, just recreating the PR here on master instead of another side branch.


ekosmin:
Yes, so I realized! The issue was that I wasn't setting the new `numTrials`variable in the test's state. I forgot that `numTrials` wasn't defaulting to the length of the `trials` array as it is during normal loading.

sfentress:
:+1:

PR #132, opened 2017-06-02
Clutch breeding levels

ekosmin:
This PR covers the Layout B template, i.e. the clutch breeding game. The clutch game template borrows heavily from the egg game template, but with most of the animations and interaction options stripped out.

Depending on who did most of the work on the original egg game template (@sfentress or @kswenson), it may make sense for you to trade off the code review.


src/code/components/clutch-view.js
```diff
@@ -0,0 +1,84 @@
+import React, {PropTypes} from 'react';
```
ekosmin: At the moment, a `clutch-view` is basically a `stable-view`, but with pages and interactivity. In the future, it should add more functionality, like setting grid parameters, so I broke it into it's own component here.

src/code/reducers/helpers/load-state-from-authoring.js
```diff
@@ -39,10 +39,10 @@ function processAuthoredGametes(authoredChallenge, drakes, state) {
             : state.gametes;
 }
 
-function processAuthoredDrakes(authoredChallenge, trial, template) {
+function processAuthoredDrakes(authoredChallenge, trial, template, unscrambledTrialNumber) {
   // takes authored list of named drakes ("mother", etc) and returns an
   // array specific for this template
-  const authoredDrakesArray = template.authoredDrakesToDrakeArray(authoredChallenge, trial);
+  const authoredDrakesArray = template.authoredDrakesToDrakeArray(authoredChallenge, trial, unscrambledTrialNumber);
```
ekosmin: I introduced the idea of an `unscrambledTrialNumber` into the `authoredDrakesToDrakeArray` function, to expand the flexibility of the authoring doc a little. Now, you can specify a combination of randomized and non-randomized information in the same challenge definition.

Below, you can see the way I used it in this PR: in level 16, I created 3 sets of parent drakes and matching target drakes to match the spec. I want to be able to see those trials in a random order, *but* I also want to ensure that the first trial only allows the mother's alleles to change, the second only allows manipulation of the father, and the third trial manipulates both. So, to get the correct trial information, I use the "scrambled" number to look up the drake information, and the "unscrambled" number to look up which parents can be manipulated.

src/code/reducers/index.js
```diff
@@ -31,6 +31,63 @@ function initialState() {
           });
 }
 
+function fertilize(state) {
```
ekosmin: This was a general question I had - what is the purpose of this `fertilize` function? I see that we call it in the code, but I'm not sure what it does differently than `BioLogica.breed`. You can see that I used the same logic below to breed random clutches, but I wonder if I can replace that logic with calls to `BioLogica`

src/resources/authoring/gv2.json
```diff
@@ -379,6 +379,104 @@
...
     },
+    "clutch-5drakes-starterTraits": {
+      "container": "FVContainer",
+      "template": "ClutchGame",
+      "challengeType": "create-unique",
+      "interactionType": "select-chromosomes",
+      "instructions": "Create 5 different baby drakes!",
+      "showUserDrake": true,
+      "randomizeTrials": true,
+      "changeableDrakes": [
```
ekosmin: I used `changeableDrakes` and `staticDrakes` here so that each trial could be used in multiple ways - either for a challenge where mother, father, or all genes can be manipulated

src/code/reducers/index.js
sfentress: The `fertilize` function uses `state.gametes` to create an offspring given user-created gametes.

If you look at the state during an egg game, you'll see something like

```
gametes: {
    currentGametes: [
      {
        '1': 'a',
        '2': 'b',
        XY: 'y'
      },
      {
        '1': 'b',
        '2': 'b',
        XY: 'x2'
      }
    ],
```

these are the sides that the user (or possibly the randomization function) has selected for each chromosome pair, for each gamete.

These together represent the entirety of the offspring. There is no more randomization. The annoying psrt (all this code) is to go from that list of sides to an organism. We have to get all the chromosomes, get the appropriate side, and concatenate all the alleles together. I don't believe that there is a Biologica function to do this, but it could certainly be added.

But unless I'm missing something, I would *not* use this for the clutch breeding. Instead I would just use `BioLogica.breed(org1, org2)`.

You can see an example of this in the non-Redux example `/examples/basic-examples/breeding.html` (http://concord-consortium.github.io/geniblocks/examples/basic-examples/breeding.html). I believe I added it as one of the first Redux actions way at the start, but then we removed it because nothing (yet) was using it.

ekosmin: Super helpful, thanks Sam - makes sense that `fertilize` is basically for breeding a very specific child from two parents, where we just need random children for clutch-breeding.

src/code/reducers/index.js
```diff
@@ -31,6 +31,45 @@ function initialState() {
...
+    if (chromosome && chromosome.alleles && chromosome.alleles.length) alleleString += "b:" + chromosome.alleles.join(",b:") + ",";
+  }
+
+  return state.setIn(["drakes", 2], {
+    alleleString,
+    sex
+  });
+}
+
+function breedClutch(state, clutchSize) {
```
ekosmin: @sfentress clutch breeding is much simpler now, and uses crossover! This should simplify correct move calculations

src/code/templates/clutch-game.js
```diff
@@ -495,45 +498,23 @@ export default class ClutchGame extends Component {
...
+    let fatherSpec, motherSpec;
+    if (changeableDrakeType === "mother") {
+      motherSpec = { alleles: changeableDrake.getAlleleString(), sex: BioLogica.FEMALE };
+      fatherSpec = { alleles: staticDrake.getAlleleString(), sex: BioLogica.MALE };
+    } else {
+      motherSpec = { alleles: staticDrake.getAlleleString(), sex: BioLogica.FEMALE };
+      fatherSpec = { alleles: changeableDrake.getAlleleString(), sex: BioLogica.MALE };
     }
-    return authoredDrakes;
+    return [motherSpec, fatherSpec, authoredChallenge.targetDrakes[trial]];
```
sfentress: This adds a lot of complexity to this template, and I bet that we're going to want similar things in other layouts (the general idea of having different static genes between mother and father, maybe based on trial number), and it would be a shame to reproduce all this in those ones. It also relies on the side-effect of the `authoredDrakesToDrakeArray` function setting an instance variable `changeableDrakeType`, which seems sketchy -- it seems like we could easily call things out of order and have that not get set.

I think this could maybe remain for now, but the moment we find ourselves wanting the same feature elsewhere we should replace it with something more generic.

Currently we have the idea of `userChangeableGenes` and `visibleGenes`, which is not being used here because it applies to the whole template. It seems as if it wouldn't be too hard to expand that to be applied to named drakes

e.g.

```
visibleGenes: "wings legs horns",
userChangeableGenes: {
   mother: "wings, legs",
   father: ""
}
```

and then expanded further to be applied to trial number:

```
visibleGenes: "wings legs horns",
userChangeableGenes: [{
   mother: "wings, legs",
   father: ""
  },
  {
   mother: "",
   father: "horns"
  }
]
```

it would still need to be implemented at the template-level, because only the templates are aware of the names, but I think it would be a one-line addition to the templates that cared (i.e. instead of passing `userChangeableGenes` to the genomeView, they pass `userChangeableGenes[name]` (they would have already been passed in the correct trial in the array as props).

src/code/components/clutch-view.js
```diff
@@ -40,48 +29,49 @@ const ClutchView = React.createClass({
...
+      const numPages = nextProps .orgs.length / 8;
+      this.setState({page: numPages});
     }
   },
 
   render() {
-    let displayStyle = {size: this.props.height*(5/8), top: -153, marginLeft: -12},
-        stableDrakeViews = this.props.orgs.map((org, index) => {
+    let displayStyle = {size: 136, top: -153, marginLeft: -12},
+        drakeViews = this.props.orgs.map((org, index) => {
```
sfentress: Instead of creating all the views, and then slicing them, why not slice the drake list first and then create the views?

```
let firstDrake = (this.props.orgs.length - 8) - (this.state.pagesBack * 8),
    visibleDrakes = this.props.orgs.slice(firstDrake, firstDrake + 8),
    drakeViews = visibleDrakes.map(....
```

it probably doesn't make a huge amount of difference in terms of speed, because if you don't add the views then no DOM elements will be created (though you do create the views, so if you have hundreds of offspring it might become noticeable), but it just seems to me to be more easily understandable.

sfentress:
As a side-effect of all the deletions and changing lines multiple times, there are 4-5 comments hidden under "outdated" folds, but they aren't actually outdated, I believe. 😉

src/code/components/clutch-view.js
ekosmin: The stable does have some different functionality from a clutch view - it displays the # of drakes instead of the # of pages, it displays static drakes instead of hatching drakes. I don't see a way to cleanly combine them, but please let me know if that's the approach you would take

src/code/reducers/helpers/load-state-from-authoring.js
ekosmin: You're right - the variable I added should just be `trialNumber`. It's the *current* variable named `trial` that should change. I renamed it to `authoredTrialNumber` to distinguish it from the `trial` in the state, and show that it is instead an index into the authored trials.

src/code/templates/clutch-game.js
ekosmin: I like this suggestion! Much cleaner than the changes I made here, and give what you pointed out during the demo yesterday (i.e. my levels aren't as flexible as I thought), I'm going to make this change

src/code/reducers/index.js
ekosmin: Yes, confirmed!

src/code/components/clutch-view.js
ekosmin: I fixed this now, by going back and editing earlier commits. When you look for the changes though, I suppose you'll just look at the final version? Normally I would make a "code review" commit, which is ugly, but does have the advantage of being easy for you to review.

src/code/templates/clutch-game.js
```diff
@@ -143,15 +142,21 @@ export default class ClutchGame extends Component {
     const parentGenomeClass = classNames('parent');
 
     function parentGenomeView(sex) {
+      let parentChangeableGenes;
+      // Changeable genes are either of the form ["wings, legs"] or [{mother: "wings", father:""}, {mother: "", father: "wings"]
```
ekosmin: I took a go at your suggestion changes, @sfentress! I got rid of the confusing `changeable` and `staticDrake` arrays, and replaced it with an expanded `userChangeableGenes` object. I decided to let you define the `userChangeableGenes` in either the new way or the old way (just a flat array). The resulting code that handles either form is a little messy, what do you think of these changes?

ekosmin: I put everything in a new commit on the branch, so it should be easier to review

PR #133, opened 2017-06-02
Making game usable on Safari and mobile browsers

ekosmin:
Here are the style changes I made to make Geniverse work on iPads for the classroom test last week, which coincidentally fixed Safari use. A lot of the changes required copying default styles from Chrome into the actual CSS.

Note that I did intentionally base this off the `clutchLevels` branch - since I want to finish up that PR before merging this in, I'll handle all the merging into master once the PRs are done.

### DIFF
--- a/src/code/fv-components/fv-gene-label.js
+++ b/src/code/fv-components/fv-gene-label.js
@@ -13,6 +13,10 @@ const Form = React.createClass({
     // render :: a -> ReactElement
     render: function(){
         return <SimpleSelect defaultValue={this.props.defaultOption} onValueChange={this.props.handleChange} options = {this.props.options} hideResetButton={true} editable={false} disabled={false}></SimpleSelect>;
+    },
+
+    componentDidMount: function() {
+      document.querySelectorAll("input").forEach(input => input.setAttribute('readonly', true));
     }
 });
 
--- a/src/stylus/fablevision/fv-gene-label.styl
+++ b/src/stylus/fablevision/fv-gene-label.styl
@@ -71,6 +71,21 @@
     justify-content: flex-start
     padding-right: 30px
 
+  &.fv-layout-a .geniblocks.chromosome-container .items.rtl .labels .fv-gene-label
+    right: 1490px
+
+  &.fv-layout-d .geniblocks.chromosome-container .items.rtl .labels .fv-gene-label
+    right: 235px
+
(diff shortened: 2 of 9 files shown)


src/code/fv-components/fv-gene-label.js
```diff
@@ -13,6 +13,10 @@ const Form = React.createClass({
     // render :: a -> ReactElement
     render: function(){
         return <SimpleSelect defaultValue={this.props.defaultOption} onValueChange={this.props.handleChange} options = {this.props.options} hideResetButton={true} editable={false} disabled={false}></SimpleSelect>;
+    },
+
```
ekosmin: Previously, the dropdowns were inputs, which meant every time one was clicked, a keyboard would pop up on the iPad. I disabled the dropdowns to prevent that, but it would be nice if we could keep them enabled on desktop. Perhaps I'll add a ticket for mobile-specific behavior

src/stylus/fablevision/fv-gene-label.styl
```diff
@@ -71,6 +71,21 @@
     justify-content: flex-start
     padding-right: 30px
 
+  &.fv-layout-a .geniblocks.chromosome-container .items.rtl .labels .fv-gene-label
```
ekosmin: This is pretty ugly - I really had to force some labels into the correct places in different ways on different layouts

src/code/fv-components/fv-gene-label.js
```diff
@@ -13,6 +13,10 @@ const Form = React.createClass({
...
+    componentDidMount: function() {
+      document.querySelectorAll("input").forEach(input => input.disabled = true);
```
sfentress: This seems to have the side-effect that it won't notice you clicking away from the pulldowns to dismiss them, because they are no longer focused and so no longer blur.

I tested this out instead with 

```
document.querySelectorAll("input").forEach(input.setAttribute('readonly', true));
```

and it seemed to work well on my Android phone, as far as I could see, but you would need to do more testing of it.

ekosmin: Yes, as you suggested, I did find a better library after the fact, where "searchable" is a configurable parameter. Something something hindsight...

https://jedwatson.github.io/react-select/

Let me give that snippet a try!

ekosmin: LGTM, merging with this change

PR #134, opened 2017-06-05
HUD notifications

ekosmin:
This PR moves all ITS and popup messages into the HUD, spoken by a character - the old messages are no longer shown. I removed the old popup dialogs in the simplest way possible: I display them in the HUD, and append any ITS feedback when it comes in. Some old dialogs used the "ok" button to advance the challenge, so I mapped all of those buttons to the "close" button of the dialog instead.

Let me know how it looks, I'm happy to discuss any of it more, thanks @sfentress! I also based this off the `mobileBrowser` branch, and will merge it in once that PR wraps up.


src/code/actions.js
```diff
@@ -409,7 +409,8 @@ function _acceptEggInBasket(eggDrakeIndex, basketIndex) {
   return {
     type: actionTypes.EGG_ACCEPTED,
     eggDrakeIndex,
-    basketIndex
+    basketIndex,
+    meta: {sound: 'receiveCoin'}
```
ekosmin: Now playing a sound when an egg is accepted, until some kind of visual cue can be shown.

src/code/app.js
```diff
@@ -106,8 +104,6 @@ function renderApp() {
                           <Route path="/(:level/:mission/:challenge)" component={ChallengeContainerSelector} />
                           <Route path="/(:challengeId)" component={ChallengeContainerSelector} />
                         </Router>
-                        <ModalMessageContainer />
-                        <NotificationContainer />
                       </div>;
```
ekosmin: I moved the messages and notifications into the FV container so they're scaled properly

src/code/components/modal-alert.js
```diff
@@ -86,21 +87,32 @@ class ModalAlert extends React.Component {
...
-        <div style={dialogStyle(this.props.top)} >
-          <h4 id='modal-label'>{t(this.props.message)}</h4>
-          {awardView}
-          {explanationView}
-          {leftButton} {rightButton}
-        </div>
-      </Modal>
-    );
+    if (this.props.show) {
+      // Only trial-end and next-trial modals are shown
```
ekosmin: I'm still using `modal-alerts` to show information that don't belong in the HUD scroll. For now, that just includes the "next level" button and the end of level feedback

src/code/components/notifications.js
```diff
@@ -8,27 +8,44 @@ class Notifications extends React.Component {
...
+      onCloseNotifications();
+    };
+  }
 
+  render() {
+    const speaker = this.props.messages.length > 0 ? <div className="fv-character"></div> : null,
+          message = this.props.messages.length > 0 
+                    ? <div className="notification">
+                        <div className="close-button" onClick={ this.handleClose(this.props.onCloseButton, this.props.onCloseNotifications) }></div>
+                        { this.props.messages.length > 1
```
ekosmin: I wrote support for multi-page messages, though we don't have any of them yet

src/code/modules/notifications.js
```diff
@@ -1,26 +1,58 @@
...
   switch (action.type) {
     case GUIDE_MESSAGE_RECEIVED:
       if (action.data) {
-        return state.concat(action.data.message.asString());
+        let currentMessage = state.messages.length > 0 ? state.messages[0] + " " : "";
+        return Object.assign({}, state, {messages: [currentMessage + action.data.message.asString()]});
       }
       else
         return state;
+    case actionTypes.MODAL_DIALOG_SHOWN:
```
ekosmin: This is a little bit of a hack - old modal dialog messages are now handled here for display. Really, the actions should be renamed for clarity, since we no longer show a modal dialog at all

src/code/reducers/modal-dialog.js
```diff
@@ -18,15 +18,19 @@ const defaultRightButton = {
...
-      return state.merge({
-        show: true,
-        message: action.message,
-        explanation: action.explanation,
-        rightButton: action.rightButton || defaultRightButton,
-        leftButton: action.leftButton,
-        showAward: action.showAward,
-        top: action.top
-      });
+      if (action.showAward || (action.rightButton && action.rightButton.action === "advanceTrial")) {
```
ekosmin: Again, this is a little bit of a hack so that the modal dialog doesn't handle messages that `notifications` are listening for. The actions should be reworked

ekosmin:
@sfentress I've never tried this strategy, I like it a lot! Normally I review the entire changeset at once, but that *does* make things very difficult to follow. Let me try to rebase some of these commits in a "one logical change per commit" style.

I also like that article's suggestion to amend commits, rather than adding additional patches down the line. Hopefully I don't break the branch figuring this out :)

ekosmin:
@sfentress, I did some editing and squelching, let me know what you think now! Should be much more readable by commit now

src/code/components/notifications.js
```diff
@@ -8,27 +8,34 @@ class Notifications extends React.Component {
...
   }
 
   render() {
-    const { location } = this.props,
-          style = location ? { left: location.left, top: location.top, right: 'auto' } : null,
-          messages = this.props.notifications.map((message, i) =>
-                        <div key={i} className="notification">{ t(message) }</div>);
-
+    const speaker = this.props.notifications.length > 0 ? <div className="fv-character"></div> : null,
+          message = this.props.notifications.length > 0
```
sfentress: Not a big deal, but what would you think of just returning early with `null` (or whatever the correct idiomatic React is) first if `this.props.notifications.length === 0`? It would simplify this (it's not clear on first glance that every piece of this view is conditional on `length > 0`) and mean that you don't need to render anything at all.

ekosmin:
Alright, I took your suggestion and clarified some of the logic by dispatching two actions instead of one. Tests got more complicated, but I think the code looks nicer! Is this what you had in mind?

sfentress:
I think this is fine. The only thing that was still tripping me up was that the only time we send a message to `notifications`, we also send it to `modal-dialog`, so I was confused as to why they weren't always showing the same message. I finally realized that, in the reducer for `modal-dialog`, it's checking to see if it's showing an award or has an advance button, and otherwise doesn't change the state.

In my mind it would be clearer if the action selectively sent a `MODAL_DIALOG_SHOWN` action only if the modal dialog should actually be shown. But I can see it either way, and I'm having trouble wrapping my head around this anyway, so it may also just be my confusion.

If you think it would make sense to be more selective about the `MODAL_DIALOG_SHOWN`, then that could be good. or if you think it's good like this, feel free to merge. :+1:

sfentress:
One thing I noticed is that the explanations are not given by the characters, so they are never seen now. But maybe  that's just because the branch is in a weird state without the ITS sending messages?

screen shot 2017-06-12 at 10 47 21 am

ekosmin:
Thanks for taking another look @sfentress! I ended up taking your suggestion to put the logic where the action is dispatched instead of on the other side - it makes it easier to understand the different use cases this way. 

Per yoru comment about the ITS messages, they don't currently work unless you have the `student` params in the URL. A link like this should work:

http://models-resources.s3-website-us-east-1.amazonaws.com/geniblocks/branch/hudNotifications/?class_info_url=https%3A%2F%2Flearn.staging.concord.org%2Fapi%2Fv1%2Fclasses%2F128&domain=https%3A%2F%2Flearn.staging.concord.org%2F&domain_uid=482&externalId=798&logging=false&returnUrl=https%3A%2F%2Flearn.staging.concord.org%2Fdataservice%2Fexternal_activity_data%2F27aa34f5-190b-4295-8784-6a3a41a47647#/2/1/1

PR #136, opened 2017-06-13
End mission dialog

ekosmin:
Last PR for now @sfentress, if you have a chance to take a look. This covers the new end of mission dialog, which includes a gem summary of your completed mission.

I tried my best to clean up the two commits to make them reviewable sequentially, but here it might still be simpler to do it all at once. The second commit is a refactor of the "gem history" into its own component, which ended up consuming all the logic in the "gem" file from the first commit. Perhaps this is an argument for squashing the commits together entirely.


src/code/components/modal-alert.js
```diff
@@ -1,65 +1,21 @@
 /*
  * Based on ReactOverlays demo at http://react-bootstrap.github.io/react-overlays/examples/#modals
  */
-import Modal from 'react-overlays/lib/Modal';
-import Button from './button';
-import ChallengeAwardView from './challenge-award';
+import EndLevelDialogView from '../fv-components/end-level-dialog';
 import React, { PropTypes } from 'react';
-import t from '../utilities/translate';
```
ekosmin: Since `modal-alert` no longer displays user messages, a lot of its functionality can be removed. Really, these deletions belonged in the `hudNotifications` PR.

src/code/fv-components/end-level-dialog.js
```diff
@@ -0,0 +1,79 @@
...
+    <div className="end-level-dialog-container">
+      <div className="end-level-dialog-backdrop"></div>
+      <div className="end-level-dialog-background"></div>
+      <div className="end-level-dialog-overlay"></div>
+      <div className="end-level-dialog-screen">
+        <div className="challenge-complete-text">
+          /{t("~VENTURE.END_LEVEL")}
+        </div>
+        <div className="end-level-flavor-text">
+          {"Geni_retrieve <get> boot_all </get>"}
```
ekosmin: Lots of "flavor text" I took out of FV screenshots here...I figured it's mostly gibberish so it wouldn't need to be localized later.

src/code/fv-components/end-level-dialog.js
```diff
@@ -0,0 +1,79 @@
...
+              {t("~BUTTON.TRY_AGAIN")}
+            </div>
+          </div>
+          <div className="continue-button-container">
+            {t("~BUTTON.CONTINUE")}
+            <div className="continue-button" onClick={onNextChallenge}></div>
+          </div>
+        </div>
+      </div>
+    </div>
```
ekosmin: The structure of this returned `div` is rather large - I wonder if it would be cleaner for me to break it into a few different private components.

src/code/fv-components/gem-set.js
```diff
@@ -0,0 +1,63 @@
+import React, {PropTypes} from 'react';
+import { getChallengeScores } from '../reducers/helpers/challenge-progress';
+
+export const GemView = ({challengeNum, score, highlight}) => {
```
ekosmin: I followed the pattern of `BasketSetView` here, which adds `BasketView` as an internal component

src/code/reducers/helpers/challenge-progress.js
```diff
@@ -36,4 +36,23 @@ export function getScoreValue(score){
     default:
       return score >= 2 ? scoreValues.BRONZE : scoreValues.NONE;
   }
-}
\ No newline at end of file
+}
+
+export function getChallengeScores(level, mission, challengeCount, progress) {
```
ekosmin: This logic came from `challenge-award.js`, which was using it to generate the old end-mission dialog. It's used in two different components now, so I figured it would be better off in a utility class here.

src/code/fv-components/end-level-dialog.js
sfentress: I think it's ok. It's a self-contained component. I wouldn't break it up just for the sake of breaking it up, until we wanted to reuse the parts.

src/code/fv-components/gem-set.js
```diff
@@ -0,0 +1,63 @@
...
+    }
+    return progressImages;
+  };
+
+  let challengeScores = getChallengeScores(level, mission, challengeCount, progress),
+      progressImages = [];
+
+  for (let challengeNum = 0; challengeNum < challengeCount; challengeNum++) {
+    let gemNumber = parseInt(challengeNum) + 1;
+    progressImages = addAwardImage(progressImages, challengeCount, gemNumber, challengeScores[challengeNum], challenge);
```
sfentress: This is minor, but it feels odd to keep setting `progressImages` to a new array within a loop. More idiomatic would be something like
```
for (let challengeNum = 0; challengeNum < challengeCount; challengeNum++) {
  let highlight = challengeNum === currChallenge;
  progressImages.push(getAwardImage(challengeScores[challengeNum], highlight));
}
```
(Other minor simplifications: `parseInt` seems unnecessary; calculating the `highlight` to make `getAwardImage` more semantic; and `pieces` doesn't seem to be used in `addAwardImage`).

src/code/utilities/lang/en-us.js
```diff
@@ -62,5 +62,12 @@ export default {
...
   // Messages from ITS
-  "~ITS.GREETING": "Hi there user!"
+  "~ITS.GREETING": "Hi there user!",
+
+  // VenturePad
+  "~VENTURE.END_LEVEL": "CHALLENGE COMPLETE",
+  "~VENTURE.YOU_EARNED": "You've earned",
+  "~VENTURE.AMBER_GEM": "Amber Crystal",
+  "~VENTURE.RED_GEM": "Red Crystal",
+  "~VENTURE.BLUE_GEM": "Blue Crystal"
```
sfentress: Another minor comment: I like how the css classes just use `.gem-fill-0`, `.gem-fill-1` etc. It seems like it would be nice to do the same thing with the award string keys: `~VENTURE.AWARD_FIRST`, `~VENTURE.AWARD_SECOND` or something. That way, if/when the GV team decides that they want to reorder to colors, or use different colors, you just need to change the CSS file and the Strings file, instead of also changing the keys everywhere in the code.

Again, not really important.

src/code/fv-components/gem-set.js
ekosmin: Good points all around! Your suggested changes make the functionality much clearer. Is that what you mean by more semantic - the parameters are more descriptive of the functionality?

I'll make those updates and merge. Again, I really appreciate you taking such a quick and thorough look through these PRs :)

PR #137, opened 2017-06-15
Updating Gem HUD Display

ekosmin:
Hi @sfentress! Tossing this to you since we'll be working on this stuff together - happy to talk about it tomorrow morning. These three commits should be easy to review sequentially.

This PR updates the gem images, adds progress tracking to the HUD, and changes the way gem saving works. There were a couple of tricky state decisions to make, so I'd appreciate a look.


src/code/fv-components/bottom-hud.js
```diff
@@ -2,31 +2,57 @@ import React, {PropTypes} from 'react';
...
+    const { routeSpec: prevRouteSpec } = this.props,
+          { level: prevLevel, mission: prevMission, challenge: prevChallenge } = prevRouteSpec,
+          { routeSpec: nextRouteSpec } = nextProps,
+          { level: nextLevel, mission: nextMission, challenge: nextChallenge } = nextRouteSpec,
+          newChallenge = (prevLevel !== nextLevel) || (prevMission !== nextMission) || (prevChallenge !== nextChallenge);
+
+    // Only update the visible gems on a new level, or whenever a level finishes
+    if (newChallenge || this.currentProgress === {} || nextProps.showAward) {
+      this.setState({currentProgress: nextProps.challengeProgress});
+    }
```
ekosmin: This is the fix for the bug that popped up during the demo, where gems would update mid-trial. Now, the gems in the HUD only change when a new level starts, or when a level finishes.

src/code/fv-components/end-level-dialog.js
```diff
@@ -19,10 +19,9 @@ const EndLevelDialogView = ({challengeAwards, onNextChallenge, onTryAgain}) => {
       mission = challengeAwards.routeSpec.mission,
       challenge = challengeAwards.routeSpec.challenge,
       challengeCount = challengeAwards.challengeCount,
-      progress = challengeAwards.progress;
+      progress = challengeAwards.progress,
+      currentScore = getScoreValue(challengeAwards.currentScore);
```
ekosmin: This is the fix for showing the correct gem for a completed challenge - even if it is worse than the "better" gem saved in your progress. The solution was to calculate the current score based off your current moves, rather than use the score in your progress, which doesn't change if your new attempt is worse.

sfentress:
I'm going to be unpopular and suggest a big change, but w can talk about it when we meet and optionally ignore it completely.

I think that the `challengeProgress` is a hard-to-understand property, it was kind of tacked on right at the start of trying to keep track of score, and it adds complexity because (a) it can change in the middle of a level, at the end, or not at all, and (b) we have to re-process it every time we want to read our state. Our processing is also universal across all challenges, so we always have to interpret "0" and a blue diamond, or whatever.

If we were imagining the Redux state from scratch, we would assume that there would be some state that corresponded to the jewels you've received. This would only change when you received a jewel. The BottomHUD would then only need to watch this property.

Also, if we're thinking about saving a "top-level" state to the backend, like what jewels you've earned, but not saving your momentary state of what you're doing right now in the level, it's clear that we're going to get messed up by this progress object that sometimes changes during a trial.

I think we're going to want to create a new top-level state property called "score" or something, that tracks your final score on each challenge. This would be set by a reducer when you end a challenge -- and we should simplify the end of challenges by having all challenges ended by the same action. Currently, we end challenges in a kind of weird way: a whole bunch of different possibly-game-ending actions call

```
let progress = updateProgress(state, action.correct);

return state.merge({
  success: action.correct,
  challengeProgress: progress
});
```

and then they also may call

```
if (args.isChallengeComplete)
      dispatch(showCompleteChallengeDialog());
```

which is kind of used as a proxy to mean "end the challenge." Instead, I think we should have an explicit "endChallenge" action, which sets the score and shows the dialog.

The score state would end up being large because it's a nested array of levels and challenges, but at base it's simply a list of the score for each challenge -- actually, maybe a tuple of your last score and your highest score, so we can show both jewels on the screen at the same time. Then we never need to process this list, we can simply read it to know what jeweles to show, whether in the end-of-challenge popup, the bottomHUD, or on a list of every single level. And we can save this as-is to the backend.

Thoughts? Let's discuss by voice.

sfentress:
Recapping our discussion: This looks good, and we'll look into making the changes to the gems in another PR.

This PR is :+1: after rebasing.

Summarizing the changes (if I recall correctly), we'll have two new state props, `currentGem` and `gems`, and we'll get rid of `progress`.

`currentGem` will be updated by the reducers for each action, and will reflect the gem you'd get if you won the challenge then and there. So it starts at zero and it will go up if (a) you submit an incorrect egg in the sorting game, (2) you move count goes above your goal moves, (3) other mistakes per individual templates.

We'll still have `moves` and `score`, as these are user-facing numbers in certain templates.

On winning a challenge, your `currentGem` moves into the `gems` hash if and only if it is better than the existing gem, if there is one.

`gems` will eventually be saved to the backend, `currentGem` will not.

Does that all sound like what we discussed?

sfentress:
...oh, and all the actions related to ending a challenge will get refactored into a glorious, seamless, easy to understand generic pattern for ending a challenge.

ekosmin:
So glorious!

That does sound like what we discussed. I think we're on the same page, I wanted to confirm: `currentGem` will only ever have values from -1 to 2, to represent the different gem choices. Actions like `submitDrake` and `submitEgg` will update `currentGem` based on other information in the state like `score` or `moves`. Does that sound right?

sfentress:
That sounds right, but what does -1 mean in this case? I'd assume you start at 0 immediately when starting a challenge, since you've made no errors. Or is that if you go beyond 2, it loops down to -1? That could be possible, but it seems that we could also just keep counting up the errors (3+) because a) it could be good for logging, b) maybe one day they'll change the scoring, and we could still just not show gems for 3+.

ekosmin:
Right, -1 would be "no gem" - we should clarify if earning no gem on a level is different in any way from never attempting the level in the first place. In my mind, we do loop from 2 to -1. Tracking the number of mistakes seems useful, but we should keep in mind that a) errors mean slightly different things for different levels, b) even though right now 1 error == `gemScore = 1`, that relationship probably won't hold as scoring functions change. If we want to keep track of a number of errors, we might want to do it in another way.

PR #139, opened 2017-06-16
Zoom template

sfentress:
Hi @ekosmin, this adds a very simple template for embedding an iframe with an arbitrary url in the "zoom room."


src/code/reducers/helpers/load-state-from-authoring.js
```diff
@@ -117,7 +117,8 @@ export function loadStateFromAuthoring(state, authoring, progress={}) {
         numTrials = authoredChallenge.numTrials || (trials ? trials.length : 1),
         trialOrder = createTrialOrder(trial, trials, state.trialOrder, authoredChallenge.randomizeTrials),
         drakes = processAuthoredDrakes(authoredChallenge, trialOrder[trial], template, trial),
-        gametes = processAuthoredGametes(authoredChallenge, drakes, state);
+        gametes = processAuthoredGametes(authoredChallenge, drakes, state),
+        zoomUrl = authoredChallenge.zoomUrl || null;
```
ekosmin: Looks like other undefined pieces of the authoring doc stay `undefined` rather than `null` - what's the reasoning for setting the `zoomUrl` to `null` by default?

src/code/reducers/index.js
```diff
@@ -134,6 +134,14 @@ export default function reducer(state, action) {
         correct: state.correct + 1
       });
     }
+    case actionTypes.ZOOM_CHALLENGE_WON: {
+      let progress = updateProgress(state, true);
```
ekosmin: I don't think the second parameter is necessary here, but probably not worth changing since this is all getting refactored soon anyway

src/code/reducers/helpers/load-state-from-authoring.js
sfentress: Totally true, fixed it.

I don't know at this point whether it would also be worth removing props from the state tree if they don't exist, instead of leaving them cluttered there, but not sure whether it's worth it.

src/code/reducers/index.js
sfentress: I was thinking that, but left it partly to be the same as the other ones that imply (without actually doing anything) a victory, to make refactoring clearer. Or, at least, that was my vague  half-thought.

sfentress:
I made a comment about removing all undefined/null authored props from the state being unnecessary, but then I realized it was basically a one-line change. Can you take a look and see if 58ff553 makes sense to you, or might cause unexpected problems?

ekosmin:
That change looks really nice! It was a pain to include the empty pieces of state in every test, so this seems useful going forward

PR #140, opened 2017-06-16
Fixing all console errors

ekosmin:
There were a handful of React errors that were thrown in every game, which made catching actual errors much more difficult. I resolved every error I could find in this PR!

### DIFF
--- a/src/code/reducers/helpers/load-state-from-authoring.js
+++ b/src/code/reducers/helpers/load-state-from-authoring.js
@@ -105,10 +105,24 @@ export function loadStateFromAuthoring(state, authoring, progress={}) {
         template = templateName && templates[templateName];
   if (!template) return state;
 
+  let userChangeableGenes = authoredChallenge.userChangeableGenes;
+  // Changeable genes are either authored in the form "wings, legs" or [{mother: "wings", father:""}, {mother: "", father: "wings"}]
+  if (typeof userChangeableGenes === "object") { 
+    let userChangeableGenesByTrial = [];
+    for (let trial = 0; trial < userChangeableGenes.length; trial++) {
+      userChangeableGenesByTrial.push({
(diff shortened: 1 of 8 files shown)


src/code/reducers/helpers/load-state-from-authoring.js
```diff
@@ -105,10 +105,24 @@ export function loadStateFromAuthoring(state, authoring, progress={}) {
         template = templateName && templates[templateName];
   if (!template) return state;
 
+  let userChangeableGenes = authoredChallenge.userChangeableGenes;
```
ekosmin: This was the only tricky bit - until, the new `userChangeableGenes` format didn't have its authored gene string split into an array, like the old format did. Now the behavior matches for both formats. I got kind of lucky that this even worked before...

sfentress: That seems ok. When I was initially thinking about `userChangeableGenes` (or even `visibleGenes`, which isn't covered here) having optional trials and named drakes, I was thinking that the names would be up to the templates, instead of hardcoding like it is here. For instance, if one of the intro challenges (one drake visible) wanted to have different genes editable on different trials, it would have to decide that the one drake was called "mother" now.

I think the way to do that would be to update the `split` function to recursively search down the tree and just convert any strings it finds.

That said, in the spirit of YAGNI (you ain't gonna need it), this is all we need now and in all likelihood all we're going to need, so I think this is fine.

ekosmin: OK, this is a good point - it's needlessly restrictive to force the names of the fields like this. I agree with leaving it for now, but when we do end up addressing `visibleGenes` (where a change like this is more likely to get used), it should be worth updating `userChangeableGenes` as well.

PR #141, opened 2017-06-19
Update its offspring message

sfentress:
Quick PR to update the ITS message for submitted bred drakes.

This involved adding to the action, which is always annoying for things that only matter for logging/ITS, but seems established at this point. Let me know if you think of another way to do this.


test/actions/submit-drake.js
```diff
@@ -23,32 +23,28 @@ const correctCharacteristics = {
...
       getState = () => ({
         routeSpec: {level: 0,mission: 0, challenge: 0},
         drakes: [
           {
-            phenotype: {characteristics: correctCharacteristics},
             alleleString: correctAlleles,
             sex: 0
           },
           {
-            phenotype: {characteristics: userCharacteristics},
```
sfentress: I don't think `phenotype` is ever a part of the drakes array, so not sure why these need to be here. Maybe something in the past that was refactored. I removed these from the test, and it seems to pass fine.

test/actions/submit-drake.js
```diff
@@ -162,6 +158,41 @@ describe('submitDrake action', () => {
...
+
+    actions.submitDrake(0, 1, correct, null, 2, 3)(dispatch, getState);
+
+    it('should call dispatch with the correct _submitDrake action', () => {
+      var submitDrakeAction = dispatch.calls[0].arguments[0];
+      expect(submitDrakeAction).toEqual({
+        type: types.DRAKE_SUBMITTED,
+        species: "Drake",
+        challengeCriteria: {
+          phenotype: submitDrakeAction.challengeCriteria.phenotype,  // we check valid phenotype below
```
ekosmin: Tiny thing - the comment about phenotype-checking comes from the call to `assertMatchingPhenotype` in other tests, which isn't included here. Since the point of the test is really to check that the offspring pieces of this action are correct, the assertion might be unnecessary, but the comment should probably reflect that.

kswenson:
Long ago we discussed the idea that in theory the ITS middleware could be responsible for adding additional information that ITS required to the action so that the client templates wouldn't be responsible for doing so, but the refactoring that would be required would presumably be prohibitive right now.

sfentress:
@ekosmin :+1: to comment change.

@kswenson I do think that might be better, although one of the added complications is things like templates having their own knowledge of things like who a bred drakes parents are, and there wouldn't be an easy way for the ITS middleware to figure that out after the fact. And we can't just say "let's standardize template drake arrays" because as soon as we introduce something like backcrossing, we won't be able to say "mother is always at index 0" or whatever.

This could possibly be resolved by having this information in the actions that *do* care, as we do now (parent indices in `breed`, in this case); always sending these actions to ITS; and having ITS be prepared to look in its history for knowledge of who a drake's parents were. Unfortunately, the ITS system isn't planning on doing that, though, so each action needs to have all the possible historic information the ITS needs.

sfentress:
Oh shoot, I just realized this whole thing doesn't work...

>  And we can't just say "let's standardize template drake arrays" because as soon as we introduce something like backcrossing, we won't be able to say "mother is always at index 0" or whatever.

We already basically have this problem. 

As it is now, we grab the current parents on Submit. But what if a user breeds a clutch, changes a parent, breeds another clutch, and then submits a drake from the *first* clutch? We'll send the wrong parent.

Two possible solutions:

1. Every drake keeps a record of its parents. This will need to be kept in the Redux state
2. Clear the previous clutches on parent change

Other thoughts?

2 is a much more drastic change for such an edge-case, but it's what we did for GV 1.0, and I think the way it is is confusing. And if a student submitted a drake from a previous clutch, but the parents are now different, is the ITS hint going to be useful anyway?

kswenson:
@ekosmin Everything that the ITS needs is included in or derivable from the state, which is available to the ITS middleware. A mapping from challenge and action to a list of ITS-desired state fields should be sufficient to prevent the ITS middleware from having to figure out such things on the fly. But all of this is academic until there's sufficient time/desire to tackle the refactoring required.

sfentress:
I'm going to pull this in, since we'll have another PR soon to clear old clutches, will will solve this specific edge-case.

I think it's quite possible that we'll eventually want to keep track of parents of all drakes (probably by ensuring that all old parents are in the drake list, even if they are not visible, and then we can just refer to them by id), but we'll do that when it's needed.

PR #145, opened 2017-06-22
Rooms

sfentress:
This adds rooms, which are visible before every challenge, with hotspots that are used to launch the challenge itself.

Dialog is not part of this PR.

The sketchiest part of this PR may be the new `showRoom` prop, and how it is set to `true` when authoring is loaded. Other ideas welcome, or we can leave it like this.


sfentress:
D'oh, need to fix tests.

src/code/containers/fv-challenge-container.js
```diff
@@ -30,8 +30,7 @@ class FVChallengeContainer extends Component {
 
     return (
       <div id="challenges" className={bgClasses} style={style}>
-        // TODO: put location names in the authoring document
-        <TopHUDView location={"Hatchery"} onToggleMap={this.props.onToggleMap}/>
+        <TopHUDView locationName={ this.props.location.name } onToggleMap={this.props.onToggleMap}/>
```
ekosmin: Huzzah! Thank you Sam :)

src/code/fv-components/location-indicator.js
```diff
@@ -1,18 +1,18 @@
 import React, {PropTypes} from 'react';
 
-const LocationIndicatorView = ({location}) => {
+const LocationIndicatorView = ({locationName}) => {
```
sfentress: Yeah, it was a little confusing to me too, but this is what I was thinking:

1. The location stuff should just be in one state variable. At the moment it needs to know at least the `name`, for the title at top, and the `id`, for the css classes. Eventually it may need which hotspot to enable. It made sense to name this variable `location`
2. The property passed through the HUD was originally being called `location`. But now that would have been confusing. So I changed it to `locationName`, and it points to the state at `location.name`.
3. In the authoring, I felt what would make most sense in the JSON would be to specify what "room" something is in, since that's how we've been talking about it. `room` is therefore a string that can be used to look up a `location` by `id`.

Make sense or too confusing?

PR #146, opened 2017-06-22
Fix goal moves

sfentress:
This fixes the issue where moving from one challenge to another would someone cause problems with the moves counter.


sfentress:
@ekosmin Good catch on goalMoves = 0. I believe this is fixed, and added a small test (although the test is not based on the different templates).

PR #147, opened 2017-06-23
Update ITS protocol, groupId; console.log reasons

sfentress:


src/code/utilities/guide-protocol.js
```diff
@@ -40,14 +40,24 @@ GuideProtocol.Event = function(username, session, sequence, actor, action, targe
 
 GuideProtocol.Event.Channel = 'Event';
 
+GuideProtocol.Event.prototype.isMatch = function(actor, action, target) {
```
ekosmin: I don't understand what the purpose of this function is, but this all looks good otherwise!

sfentress: Nor me. I just pulled it in as-is from https://github.com/IntelliMedia/guide-server/blame/master/shared/guide-protocol.js

PR #148, opened 2017-06-24
Firebase State Saving

ekosmin:
I could have sworn I opened a PR for this yesterday, but I must not have pulled the trigger...in any case, here's what I've accomplished for state saving, with the [schema update suggested by Sam](https://gist.github.com/sfentress/b0b83103d2109ec287160511801c96e2). I didn't understand the `stateVersion` piece of the key so I left it out, but it might be worth talking over that more.

We interact with Firebase in two different places: 

1) When a session begins, we check if the user has any stored data. If so, we load it into the state. At this point, we also check if this version of the authoring document is stored in Firebase, and if not, we load it in.

2) When a challenge ends, a new piece of middleware updates the user's state in Firebase.

I'm sure there are more clever ways to have Firebase watch the `gem` state and update automatically, but this seems to get the job done. As for reviewing this PR, the first commit stores data in a flat map, so you may want to skip it, but I left it in for posterity.

### DIFF
--- a/src/code/middleware/state-save.js
+++ b/src/code/middleware/state-save.js
@@ -1,4 +1,4 @@
-import { getClassId, getUserId } from '../utilities/url-params';
+import urlParams from '../utilities/url-params';
 
 export const authoringVersionNumber = 1;
 
@@ -29,4 +29,17 @@ export function getUserQueryString() {
         userId = getUserId();
 
   return (classId && userId) ? authoringVersionNumber + "/userState/" + classId + "/" + userId : null;
+}
+
+function getClassId() {
+  return convertUrlToFirebaseKey(urlParams.class_info_url);
+}
+
+function getUserId() {
+  return convertUrlToFirebaseKey(urlParams.domain) + urlParams.domain_uid;
+}
+
+function convertUrlToFirebaseKey(url) {
+  // Convert invalid Firebase characters (inluding periods) to their ASCII equivalents
+  return encodeURIComponent(url).replace(/\./g, "%2E");
 }
--- a/src/code/middleware/state-save.js
+++ b/src/code/middleware/state-save.js
@@ -1,17 +1,29 @@
 import actionTypes from '../action-types';
-import { getReturnUrlId } from '../utilities/url-params';
+import { getClassId, getUserId } from '../utilities/url-params';
+
+export const authoringVersionNumber = 1;
 
 export default () => store => next => action => {
-  let result = next(action),
-      nextState = store.getState();
+  let result = next(action);
 
   if (action.type === actionTypes.CHALLENGE_COMPLETED) {
-      let gems = nextState.gems, 
-          update = {};
+      let userQueryString = getUserQueryString();
 
-      update[getReturnUrlId()] = gems;
-      firebase.database().ref().update(update); //eslint-disable-line
+      if (userQueryString) {
+        let nextState = store.getState(),
+            gems = nextState.gems,
+            stateUpdate = {state: gems};
+          
+        firebase.database().ref(userQueryString).update(stateUpdate); //eslint-disable-line
+      }
     }
 
   return result;
 };
+
+export function getUserQueryString() {
+  const classId = getClassId(),
+        userId = getUserId();
+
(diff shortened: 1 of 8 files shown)


src/code/middleware/state-save.js
```diff
@@ -0,0 +1,29 @@
...
+            stateUpdate = {state: gems};
+          
+        firebase.database().ref(userQueryString).update(stateUpdate); //eslint-disable-line
+      }
+    }
+
+  return result;
+};
+
+export function getUserQueryString() {
```
ekosmin: I wasn't sure where to put utilities like this - they're in the middleware for now, but we may want them somewhere else if we build out our database interactions

src/code/middleware/state-save.js
```diff
@@ -0,0 +1,29 @@
+import actionTypes from '../action-types';
+import { getClassId, getUserId } from '../utilities/url-params';
+
+export const authoringVersionNumber = 1;
```
ekosmin: I played around with the idea of dynamically figuring out which version of the authoring document is getting used, and incrementing the version number if it doesn't exist in Firebase. I couldn't figure out a way to do this without querying the root of the database and getting back all the data.

I suppose if we had a separate mapping of version numbers to authoring documents somewhere else in Firebase, we could do this more easily.

sfentress:
...to be clear about the migration description in point 2 above, it doesn't require putting support in the migrations in this version. Simply by giving the user data a version number (1, at the start), it allows us to add migrations later.

ekosmin:
Thanks for the comments @sfentress! I think I understand all of your points - I threw in `stateVersion` to make structural changes simpler in the future, even though I don't yet check it anywhere in the code. I also now check all actions for updates to the gem state (using `stringify`, let me know if you think that's a bad idea).

As for point (1), I like the idea of storing the entire authoring file in Firebase. If we wanted to create a new version, would the process look something like: first, manually create a new `gameVersion` entry in Firebase with the updated authoring doc. Then, update the `authoringVersionNumber` in the codebase, so it reads from and writes to the new location.

src/code/utilities/url-params.js
```diff
@@ -12,4 +12,16 @@ var urlParams;
...
+export function getClassId() {
+  // Pull the class number out of the class info
+  let regex = /classes\/(\d*)/g,
+      matches = regex.exec(urlParams.class_info_url);
+
+  return matches ? matches[1] : null;
+}
+
+export function getUserId() {
+  return urlParams.domain_uid;
```
sfentress: Just for my own understanding: do you know why the user_id is in the url as `domain_uid`?

Also, the classId above is pulling a number from a property called `classes` -- are we sure this won't be a list of numbers if the user is in more than one class? Will this still work correctly?

sfentress:
This all looks great. There's just one small change I think needs to be made: currently the user state is being saved as

```
state: [[0,0],[1,0], ...]
```

there is no mention that this is `gems`, so the moment we want to save anything else, we're going to need to move all that data someplace via a migration (which I guess we will be able to support, so that's ok).

Why not save it as

```
state: {
   gems: [[0,0],[1,0], ...]
}
```

as then we can add new properties without causing breaking changes?

src/code/utilities/url-params.js
ekosmin: Good questions! As for `classes`, I just signed up for 2 and it still seems to send just the desired class ID. I'm still not sure why they're named what they are, would you be able to help us out @scytacki?

ekosmin:
Got up a new commit that stores `gems` as a field within the `state`, let me know what you think!

scytacki:
Sam pinged me about this. Here is my recommendation:
Use the full class_info_url for the class key, that way it will work if there are multiple portals using GV. Likewise for the user key, combine the domain and the domain_uid into one string.

src/code/utilities/url-params.js
scytacki: The domain_uid is named that way because this is a generic feature that used for single sign on. Where the 'domain' is the SSO provider.  As I wrote below I'd suggest combining the domain and the domain_uid, so it will work if GV is used in multiple portals.

PR #149, opened 2017-06-26
Room dialog

sfentress:
This adds the pre-challenge in-room dialog.

It just uses the `character` attribute that was already in the narrative dialog to set the css class. It also adds a property called `defaultCharacter` to the location, which is passed to the narrative component for ITS messages or other messages without characters set.

I've started added the dialog. I'll add more soon, but figured this could be looked at without the rest for the moment.

http://geniventure.concord.org/branch/room-dialog/#/1/1/1


src/code/components/notifications.js
```diff
@@ -28,22 +28,35 @@ class Notifications extends React.Component {
...
+    if (!this.props.messages || !this.props.messages[0]) {
+      return null;
+    }
+
+    const message = this.props.messages[0],
+          text = message.text,
+          character = message.character || this.props.defaultCharacter,
+          speaker = <div className={`fv-character ${character}`}></div>,
+            // don't show close button if there's more narrative dialog
+          showCloseButton = (message.type && message.type !== "dialog") || !this.props.messages[1],
```
ekosmin: Ah and this should likely reference the new `notificationType` enum

src/code/modules/notifications.js
```diff
@@ -10,6 +10,10 @@ export const GUIDE_ERRORED = "Guide errored";
 export const ADVANCE_NOTIFICATIONS = "Advance notifications";
 export const CLOSE_NOTIFICATIONS = "Close notifications";
 
+export const notificationType = {
+  DIALOG: "dialog"
```
sfentress: 1. Yes, narrative might be better. I'll change that.
2. I think the other types (of which currently there seem to be two kinds, ITS-messages and our own canned feedback messages) could be named once we need to distinguish them. So if we don't want the ITS to append to an existing feedback message, for example, but we do want it to append to an existing ITS message, then we could label them. (They could also have been labeled now, granted.)

src/code/containers/notification-container.js
```diff
@@ -4,9 +4,11 @@ import * as actions from '../actions';
 import { advanceNotifications, closeNotifications } from '../modules/notifications';
 
 function mapStateToProps (state) {
+  let messages = state.notifications.messages.concat(state.dialog);
```
sfentress: D'oh, a previous version of how I did it, with a left over appendix. Will fix.

src/resources/authoring/gv2.json
```diff
@@ -1019,13 +1023,26 @@
               {
                 "name": "Challenge 1.1.1",
                 "id": "allele-5drakes-tutorial",
-                "room": "simroom"
+                "room": "simroom",
+                "dialog": [
+                  {
+                    "character": "YEUNG",
+                    "text": "Good to meet you, Cadet! I'm Professor Yeung, and I'm in charge of the Sim Room. We’re going to show you on this simulator how to breed five different kinds of drakes - with different combinations of arms, legs, wings, horns, and armor."
```
sfentress: From our voice meeting, thinking that something like the whole authoring file being localized may possibly make more sense.

PR #151, opened 2017-06-27
Mission Control

sfentress:
This adds 

1. The Mission Control screen, visible at `/#home`
2. Loading mission control by default if you navigate to `/`
3. Start-mission dialog if you navigate to home and at at the start of a mission, with no challenges solved (counting blank gems as "solving" a mission)
4. Middle-mission dialog if you enter the mission control and you have > 1 gems in the mission (for most users, this will only accessible if they load the game with saved-state, because currently there is no way to navigate to the MC except by editing the url)
5. On the Victory screen, the Continue button will now either open the Navigation if there are more challenges in a mission, or go back to MC if there aren't
6. If they go back to the MC via the Victory screen, they are presented with End-mission dialog before the next missions start-dialog.

Sorry if the PR is a little jumbled. I've tried to make it streamlined, but there are a number of different things in the same PR.

### DIFF
--- a/src/code/actions.js
+++ b/src/code/actions.js
@@ -128,27 +128,12 @@ export function navigateToNextChallenge() {
         nextChallenge = currentChallenge+1,
         challengeCountInMission = AuthoringUtils.getChallengeCount(authoring, currentLevel, currentMission);
     if (challengeCountInMission <= nextChallenge) {
-      let missionCountInLevel = AuthoringUtils.getMissionCount(authoring, currentLevel);
-      // there are not enough challenges...if the next mission exists, navigate to it
-      if (currentMission + 1 < missionCountInLevel) {
-        nextMission++;
-      } else {
-        let levelCount = AuthoringUtils.getLevelCount(authoring);
-        // otherwise, check if the next level exists
-        if (currentLevel + 1 < levelCount) {
-          nextLevel++;
-        } else {
-          // if no next level exists, loop around
-          nextLevel = 0;
-        }
-        nextMission = 0;
-      }
-      nextChallenge = 0;
+      dispatch(navigateToHome());
 
       if (endMissionUrl) {
         dispatch(navigateToStartPage(endMissionUrl));
-        return;
       }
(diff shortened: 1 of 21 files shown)


sfentress:
This is now built at http://geniventure.concord.org/branch/missioncontrol/

src/code/actions.js
```diff
@@ -95,6 +95,20 @@ function navigateToStartPage(url) {
   };
 }
 
+export function continueFromVictory() {
+  return (dispatch, getState) => {
+    const { routeSpec, authoring, gems } = getState(),
+          currentMission = routeSpec.level + "" + routeSpec.mission,
```
ekosmin: I can't think of a specific case where this would fail, but I'd think it should be something like `routeSpec.level + "." + routeSpec.mission` to avoid a weird case where "1.11" = "11.1"

src/code/reducers/modal-dialog.js
```diff
@@ -25,7 +25,8 @@ export default function modalDialog(state = initialState, action) {
     case actionTypes.TOGGLE_MAP:
       return state.merge({
         show: action.isVisible,
-        showMap: action.isVisible
+        showMap: action.isVisible,
+        showAward: false
```
ekosmin: Looks like this also sets `showAward` to `false` if the map is getting toggled off. Granted, this probably points more to a problem with the way I designed the way the dialogs get displayed...

src/code/actions.js
sfentress: That's a good point.

src/code/reducers/modal-dialog.js
sfentress: True, but we certainly wouldn't want to show award just because the user toggled the map off.

I guess we could imagine that the user was looking at the award, went to the map, and then wanted to look at the award again. But then we'd have to save some temporary state to say that they had been looking at the award. I'm not sure that it's necessary, though. I think if you toggle the "map" off, you expect the whole widget to go away.

PR #159, opened 2017-06-29
Classroom bugfixes

ekosmin:
Fixes the "stranded user" bug from today, and cleans up feedback delivery order


src/code/components/notifications.js
```diff
@@ -21,10 +21,10 @@ class Notifications extends React.Component {
 
   handleClose(onCloseButton, onCloseNotifications) {
     return () => {
+      onCloseNotifications();
```
ekosmin: Now closing notifications before any close button actions, so that narration created by the action isn't immediately cleared.

PR #172, opened 2017-07-12
Re-enabling Firebase authoring loading

ekosmin:
Previously, some authored challenges used arrays of empty objects to count trials. There's no way to represent those objects in Firebase, so those challenges now use the recently added numTrials variable instead.

After making the change, an issue arose where the its-log middleware was making breaking changes to dispatched actions.


src/code/middleware/its-log.js
```diff
@@ -162,7 +163,7 @@ function createLogEntry(loggingMetadata, action, nextState){
   delete context.type;
   delete context.meta;
 
-  context = makeMutable(context);
+  context = makeMutable(cloneDeep(context));
```
ekosmin: @kswenson @sfentress Before I made this change, I was finding the drakes in my `drakes` array had their `sex` set to `male`/`female` instead of `0`/`1`. Copying the context here instead of modifying it directly fixed the problem. It's not clear to me why this only started happening when the authoring started loading from Firebase, am I missing any obvious reasons?

src/code/templates/fv-egg-game.js
```diff
@@ -1343,7 +1343,7 @@ export default class FVEggGame extends Component {
       return authoredDrakes;
 
     // challengeType === 'match-target'
-    const targetDrakeCount = authoredChallenge.targetDrakes.length;
+    const targetDrakeCount = authoredChallenge.numTrials || authoredChallenge.targetDrakes.length;
```
sfentress: Awesome. It's nice to get rid of the `targetDrakes: [{},{},{}]`-style authoring.

src/code/middleware/its-log.js
sfentress: I don't quite understand. You were seeing the strings `"male"` and `"female"` after `makeMutable`? Can't see how that would have happened without diving deep into the `makeMutable` function.

ekosmin: Oh, I should have probably made this comment a couple of lines below - make mutable just makes the context mutable, and then it changes the values `0` and `1` into `male` and `female` in the following 3 lines. I was seeing those changes then persist through to the state.

PR #174, opened 2017-08-22
Save a timestamp of the last action to firebase

sfentress:
Every action the user takes will now save a timestamp to Firebase, in
a new property `stateMeta`, designed to hold the data that does not make
up a user's loadable state.

This data will allow the Dashboard to show what users are currently
active.

@ekosmin could you please take a look at this when you get a chance?

### DIFF
--- a/src/code/middleware/state-save.js
+++ b/src/code/middleware/state-save.js
@@ -3,16 +3,20 @@
 /**
  Our current user state:
 
- {
-   state: < saveable state (i.e. gems) >,
-   stateVersion: 1,
-   stateMeta: { lastActionTime: t },
-   itsData: < set by ITS >
- }
+  {
+    state: < saveable state (i.e. gems) >,
+    stateVersion: 1,
+    stateMeta: {
+      lastActionTime: t
+      currentChallenge: {l, m, c}
+    },
+    itsData: < set by ITS >
+  }
  */
 
 import urlParams from '../utilities/url-params';
 import actionTypes from '../action-types';
+import progressUtils from '../utilities/progress-utils';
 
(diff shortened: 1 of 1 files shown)


sfentress:
@ekosmin I added the current challenge the user is on/will be on to the metadata.

src/code/middleware/state-save.js
```diff
@@ -44,6 +52,21 @@ export default () => store => next => action => {
...
+  if (!authoring) {
+    return null;
+  } else if (routeSpec && routeSpec.level !== undefined) {
+    return {
+      level: routeSpec.level,
+      mission: routeSpec.mission,
+      challenge: routeSpec.challenge
+    };
+  } else {
+    return progressUtils.getCurrentChallengeFromGems(authoring, gems);
```
ekosmin: This is the only piece that confused me a bit, the idea that this piece of state contains either their current challenge or their next available challenge. Then again, any time they go to the home screen, they're directed to the next available challenge anyway, so this should be accurate, right?

sfentress: The idea is that the teachers will always want to see where a student is in the Report. If a student has finished a mission and goes back to the home room, then they don't have any current challenge in their state. So instead we use their gem state and work out what their next challenge is, using a pre-existing method. This *might* not end up being the one they chose to do next, as they could chose another challenge in the mission, or even a past mission, but it's our best guess as to where they are.

Alternatively, we could show what challenge they just finished (we'd have to add some stuff for that), but that seems less preferable, as it would look on the table that they are still working on that challenge (like maybe repeating it for a new gem).

Does that make sense?

PR #175, opened 2017-08-22
Save past gems

sfentress:
This updates the `gems` state to contain arrays of gems for each challenge, instead of a single gem. This means that instead of just keeping the last earned gem, we can keep the gem history for each challenge. This allows us to display this history in the dashboard.

Since we are saving `gems` to Firebase, we need to add a migration for all users that have already saved gems, or those users will revert back to the begining.

In adding a migration, it seemed to simplify things to add `stateVersion` to the `state` object itself. That allows us to make a single request to FB to get our data, with the version.

The combination of this and #174 will make our Firebase user data into:

```
{
  state: {
    < saveable state (gems) >,
    stateVersion: 2
  },
  stateMeta: { 
    lastActionTime: t 
  },
  itsData: < set by ITS >
}
```

@ekosmin Could you take a look at this? If it looks good, I'll rebase it on top of #174.

### DIFF
--- a/src/code/actions.js
+++ b/src/code/actions.js
@@ -7,6 +7,7 @@ import ProgressUtils from './utilities/progress-utils';
 import { getUserQueryString } from './middleware/state-save';
 import { notificationType } from './modules/notifications';
 import { getGemFromChallengeErrors } from './reducers/helpers/gems-helper';
+import migrate from './migrations';
 
 export { actionTypes };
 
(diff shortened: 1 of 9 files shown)


src/code/actions.js
```diff
@@ -1,3 +1,4 @@
+/* global firebase*/
```
ekosmin: Ohh didn't know you could do this, nice!

src/code/migrations/02_convert_gems_to_array.js
```diff
@@ -0,0 +1,26 @@
...
+  if (state.stateVersion < version) {
+    state.stateVersion = version;
+    return doUpdate(state);
+  }
+  return state;
+}
+
+// Previously: each mission had one number representing the last gem awarded.
+// Now: each mission has an array of all gems awarded for that mission.
+function doUpdate(state) {
```
ekosmin: Does this migration leave behind the old `stateVersion: 1` on stored data?

sfentress: On line 5 it updates the version in the state. Does that answer the question? Would it make more sense to you to put that in `index.js`?

PR #176, opened 2017-09-15
Remove gating temporarily

sfentress:
@ekosmin This is a one-line change, returning `false` from the function `isMissionLocked`. It seems to work fine, but wanted to check if you have thoughts about doing it this way.

I think in a few months we'll want to have this function back the way it was (perhaps alongside a Firebase property settable from the teacher dashboard), so I wanted to leave it easy to revert.


sfentress:
@ekosmin I added a couple of commits to fix the dialog issues that we realized would result from this. Could you take another quick look? Thanks!

PR #177, opened 2017-09-20
Its traits arrows

sfentress:
@ekosmin This adds highlighting and arrows around traits called out by the ITS.

screen shot 2017-09-20 at 12 49 59 pm


src/code/modules/notifications.js
```diff
@@ -27,22 +27,30 @@ export default function notifications(state = initialState, action) {
...
         }
 
-        let trait = data.message.args ? data.message.args.trait : null;
         let currentText = state.messages.length > 0 ? state.messages[0].text + " " : "";
         let newMessage = {
           text: currentText + data.message.asString()
         };
+
+        let trait;
+        // check reason.trait first, then message.args.trait
```
ekosmin: Is the idea that the ITS sends over information about the incorrect trait in different ways for different problem types?

sfentress: They haven't implemented `reason.trait` yet, and I'm not sure of their timeline to do so. The same info just so happens to be in `message.args.trait`, but it wasn't really designed for that (that's supposed to be for string substitution, I guess), so we shouldn't rely on it being there. So... there you have it.

PR #181, opened 2017-11-14
Implement css version of FV drop down styling images

dstrawberrygirl:
I used css instead of the supplied images to improve rendering and flexibility of layout of drop downs, especially the switch from arrows on the right to left depending on which side the label is positioned. I did not see any samples showing hover highlights for option selection so I improvised with a general grey background, though that can be changed fairly easily.

I wasn't sure how to get to a layout that used the brown background - do any of the current challenges use that layout?

The React Selectize SimpleSelect control did not seem to allow the option to resize the svg arrow that was included in the control so I added a simple arrow.

### DIFF
--- a/src/code/fv-components/fv-chromosome-image.js
+++ b/src/code/fv-components/fv-chromosome-image.js
@@ -27,6 +27,22 @@ const FVChromosomeImageView = ({small=false, empty=false, chromosomeDescriptor, 
     return null;
   }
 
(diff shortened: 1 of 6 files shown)


sfentress:
Hi Chris, this is currently failing on Travis, due to linting issues.

When you get it building, we'll be able to see it at http://geniventure.concord.org/branch/dropdown-styling

sfentress:
(yeah, it's set to be strict...)

PR #182, opened 2017-11-17
Update Clutch Game to new Layout H

sfentress:
Hi @dstrawberrygirl, this is one single ugly commit, but it couldn't really be broken up because it's basically one big view change to set everything to a new layout.

There are no logic changes, and the only changes to shared components are either non-changing (new size option in fv-components/parent-drake.js) or to a component only used by this layout (components/clutch-view.js).

This should be visible at http://geniventure.concord.org/branch/layout-h/#/4/2/1 soon.

I will be rebasing the sibling-breeding branch on top of this, and putting sibling breeding into this new layout.


sfentress:
@dstrawberrygirl This and #183 are all set now.

src/stylus/fablevision/clutch-game.styl
```diff
@@ -1,18 +1,139 @@
...
+    width: auto
+    margin-top: 23px
+
+  .geniblocks.breed-button-surround
+    background: none
+    height: auto
+    width: auto
+    margin-top: 0
+
+    .breed-button
```
dstrawberrygirl: There's a slight issue with the breed button when you click and hold - the image switches but the image appears tiled and off-center. Perhaps this is related to the different image sizes?

sfentress: :+1: I had noticed this with the other button sizes, and got new versions from FV, but forgot to update the breed button.

This should be fixed now (and visible when Travis has built it).

sfentress:
>  it might be good to centralize the color highlights for different backgrounds so its easier to switch between layout schemes.

That sounds good, but agree it doesn't belong here. Maybe whichever of us updates the chromosome images can remember to add that?

src/stylus/fablevision/clutch-game.styl
dstrawberrygirl: I think we have a case-sensitivity issue - Breed_ON vs breed_ON (and similarly for the other variants) - the built version of the code responds with a 404 for the mismatch

PR #183, opened 2017-11-21
Pick-the-parents (aka Sibling breeding)

sfentress:
This adds logic for pick-the-parents challenges, and updates the Clutch Game layout to be Layout H when the challenge type is "select-parents".

This PR is branched off `layout-h`, so it would make sense to merge #182 first.


sfentress:
This can be seen at http://geniventure.concord.org/branch/sibling-breeding/?localAuthoring=siblings#/4/2/3

PR #185, opened 2017-12-06
Load images before entering rooms

sfentress:
When the Challenge Container loads, or when the room changes, we check
to see if there are images to load. If there are, we load them while
displaying a new LoadingView, before rendering the regular page.

The images are listed in a file, because

1. We can control which images are loaded for each room,
2. We can be smarter about which ones to load, and not load tiny
images that we don't care if they appear late. This allows us to
speed up loading at the expense of having all the assets ready.

However, the problem with a list is that it needs to be kept up-to-date.
We'll see if this remains a problem.


dstrawberrygirl:
Agreed that a static list of images to preload is fine for now, perhaps in future we can choose to preload all image assets over a certain file size if that produces the desired result. 

The messages that appear on screen change a little too quickly - I would increase the time for each message a little, and perhaps randomize where in the list you are when you start to see the messages, so if you only see 3-4 messages on the first challenge then you would be more likely to see 3-4 new messages on each subsequent challenge (or perhaps shuffle the list on load to randomize the deck).

Otherwise all looks good, 👍

sfentress:
Turnings out if you're typing fast and fat-fingerdly type `git push origin preload-assets -df` instead of `... -f`, it will delete your branch and close your PR...

PR #191, opened 2018-02-08
Make moves counter count down; gem colors on moves screen; hide counter

sfentress:
Commit 1 is for this story: https://www.pivotaltracker.com/story/show/146888383

The moves counter now counts down instead of up, starting with goal moves + 2, essentially showing the number of moves you can make and still get a gem.

The color of the moves screen matches the gem you'll get if you make a correct submission right then.

The logic for showing the moves counter remains, it's only on templates that used to show moves.

Commit 2 is for this story: https://www.pivotaltracker.com/story/show/155005035

A template can optionally hide the moves counter, using an optional static function that takes a challenge-type and an interaction-type.


sfentress:
@dstrawberrygirl This has been updated with the "Too many moves" counter and is now ready.

src/code/templates/clutch-game.js
```diff
@@ -271,4 +271,11 @@ export default class ClutchGame extends Component {
     return 1;
   }
 
+  static shouldShowMovesCounter = function(challengeType) {
+    if (challengeType === "submit-parents") {
```
dstrawberrygirl: Knowing that test-cross is coming, would be good to add that to the list, unless there's a way we can we pick this up from an authoring flag?

PR #195, opened 2018-05-18
Limit animation duration to half a second

dstrawberrygirl:
I don't believe we use the spring motion anywhere else in the application - there's a tail to the spring that drags out the length of time until "onRest" is called which was delaying the hatch animation, so I've added a time limit. But I wanted to check if this approach was acceptable or whether there's a better way to achieve this result.

### DIFF
--- a/src/code/hoc/animate-spring.js
+++ b/src/code/hoc/animate-spring.js
@@ -16,16 +16,24 @@ import {Motion, spring} from 'react-motion';
     in its target state without further animation. This allows animations at rest to
     respond immediately to window resize events, for instance.
  */
-export default function animateSpring(stiffness=120, damping=25) {
+export default function animateSpring(stiffness=115, damping=18) {
 
   return function(WrappedComponent) {
     return class extends React.Component {
-      
+      componentDidMount() {
+        this.fixedAnimationDuration = window.setInterval(() => {
+          if (!this.isComplete) {
+            this.handleRest();
+          }
+          window.clearInterval(this.fixedAnimationDuration);
+        }, 500);
+      }
+
       static propTypes = {
         key: PropTypes.string,
         initialStyle: PropTypes.object.isRequired,
         targetStyle: PropTypes.object.isRequired,
-        onRest: PropTypes.func
+        onRest: PropTypes.func,
       }
 
       handleRest = (...args) => {


src/code/hoc/animate-spring.js
```diff
@@ -16,16 +16,24 @@ import {Motion, spring} from 'react-motion';
     in its target state without further animation. This allows animations at rest to
     respond immediately to window resize events, for instance.
  */
-export default function animateSpring(stiffness=120, damping=25) {
+export default function animateSpring(stiffness=115, damping=18) {
```
kswenson: FYI, these defaults were chosen because they are (or were) the default values for the `react-motion` library, so the default behavior would be the same with or without the HOC.

src/code/hoc/animate-spring.js
```diff
@@ -16,16 +16,24 @@ import {Motion, spring} from 'react-motion';
...
 
   return function(WrappedComponent) {
     return class extends React.Component {
-      
+      componentDidMount() {
+        this.fixedAnimationDuration = window.setInterval(() => {
```
kswenson: Seems like you could use `setTimeout()` since you're only interested in having the timer run once.

src/code/hoc/animate-spring.js
```diff
@@ -16,16 +16,24 @@ import {Motion, spring} from 'react-motion';
...
+          if (!this.isComplete) {
+            this.handleRest();
+          }
+          window.clearInterval(this.fixedAnimationDuration);
+        }, 500);
```
kswenson: The original goal of `animateSpring()` was to be a reusable HOC appropriate for any project using `react-motion`. I think offering the option of a fixed timeout is a nice enhancement, but my predilection would be to make this a parameter so that clients would get the default `react-motion` behavior by default, but could specify an appropriate timeout when desired. Having three default parameters would be inconvenient, so switching to an options argument, e.g. `interface ISpringOptions { stiffness?: number, damping?: number, timeout?: number }` might make the most sense.

src/code/hoc/animate-spring.js
```diff
@@ -16,16 +16,24 @@ import {Motion, spring} from 'react-motion';
...
+      }
+
       static propTypes = {
         key: PropTypes.string,
         initialStyle: PropTypes.object.isRequired,
         targetStyle: PropTypes.object.isRequired,
-        onRest: PropTypes.func
+        onRest: PropTypes.func,
```
kswenson: Extra comma not strictly required. 😉

dstrawberrygirl:
Thanks @kswenson for the review - setTimeout now in place and pulled out the configuration for timeout as suggested to make it an optional setting by clients.

PR #198, opened 2018-05-31
Add schema for authoring validation

kswenson:
Add schema for authoring validation [#157981472]
- add `validate` script to perform authoring validation against schema

There's a lot more that could be done here, but this should be a reasonable starting point. I've fleshed out the schema for the `FVGenomeChallenge`, but the other challenges are currently handled generically. Allele strings and gene lists could be validated with regular expressions, etc.

@dstrawberrygirl pointed out that Visual Studio Code can be configured to [show schema validation errors](https://code.visualstudio.com/docs/languages/json#_mapping-to-a-schema-in-the-workspace) in the editor. My configuration looks like
```
    // Associate schemas to JSON files in the current project
    "json.schemas": [{
        "fileMatch": ["gv-1.json"],
        "url": "./src/resources/authoring/authoring.schema.json"
    }],
```
If we adopted a naming convention like `.authoring.json` for all authoring files, then that could be coded into the `fileMatch`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "uuid": "^2.0.2"
   },
   "devDependencies": {
+    "ajv-cli": "^3.0.0",
     "babel-eslint": "^7.1.1",
     "babel-preset-es2015": "^6.18.0",
     "babel-preset-react": "^6.16.0",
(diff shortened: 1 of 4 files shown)


dstrawberrygirl:
I believe we could create a workspace settings file so that schema validation would apply to anyone using VS Code working on the project with no manual configuration required.

sfentress:
It would be great to either get the VS Code configuration into the readme, or to add the settings file.

As for the `.authoring.json` convention, is it not possible to match all `.json` files in `src/resources/authoring/`?

kswenson:
@sfentress There are quite a few files in there and I wasn't sure whether all of them were necessarily suitable for validation with the schema. One obvious example is the schema file itself. We certainly could move the schema elsewhere and configure it to validate against all files in the authoring folder.

PR #199, opened 2018-06-01
Level 6 Challenges 1-4

kswenson:
Initial implementation of level 6 challenges
- challenges 1-2 are visible target drake matching [#157981208]
- challenges 3-4 are hidden target drake matching [#157981248]

Fix layout issues with layout A challenges encountered in level 6 [#157981432]
- center genome layout within its borders
- fix layout of "non-standard" allele names (e.g. "tk")
- capitalize "MISSION" in bottom HUD for consistency with other labels
- fix React PropTypes warning when number of moves is exhausted

Fix ESLint warnings/errors

Note: This PR depends on the latest BioLogica PRs (0.11). Testing is likely to yield unexpected results without the latest BioLogica updates.

### DIFF
--- a/src/code/modules/tutorials.js
+++ b/src/code/modules/tutorials.js
@@ -1,6 +1,5 @@
 import Immutable from 'seamless-immutable';
 import actionTypes from '../action-types';
-// import t from '../utilities/translate';
 import tutorialSteps from './tutorial.json';
 
 const tutorialActionTypes = {
--- a/src/code/modules/tutorials.js
+++ b/src/code/modules/tutorials.js
@@ -1,6 +1,6 @@
 import Immutable from 'seamless-immutable';
 import actionTypes from '../action-types';
-import t from '../utilities/translate';
+// import t from '../utilities/translate';
 import tutorialSteps from './tutorial.json';
 
 const tutorialActionTypes = {
--- a/src/stylus/fablevision/fv-gene-label.styl
+++ b/src/stylus/fablevision/fv-gene-label.styl
@@ -1,3 +1,7 @@
+/* establish a stacking context, so all interior z-index values are local */
+.geniblocks.genome
+  z-index: 1
+
 .geniblocks.genome #mountNode .react-selectize.open
   z-index: 20
 
--- a/src/stylus/fablevision/fv-gene-label.styl
+++ b/src/stylus/fablevision/fv-gene-label.styl
@@ -1,6 +1,6 @@
 /* establish a stacking context, so all interior z-index values are local */
 .geniblocks.genome
-  z-index: 1
+  transform: scale(1)
 
 .geniblocks.genome #mountNode .react-selectize.open
   z-index: 20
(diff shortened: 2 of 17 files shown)


src/code/modules/tutorials.js
```diff
@@ -1,6 +1,6 @@
 import Immutable from 'seamless-immutable';
 import actionTypes from '../action-types';
-import t from '../utilities/translate';
+// import t from '../utilities/translate';
```
dstrawberrygirl: Commented out so can probably remove?

test/templates/clutch-game.js
```diff
@@ -1,22 +1,22 @@
 import ClutchGame from '../../src/code/templates/clutch-game';
 import GeneticsUtils from '../../src/code/utilities/genetics-utils';
-import reducer from '../../src/code/reducers/';
-import { navigateToChallenge } from '../../src/code/actions';
+// import reducer from '../../src/code/reducers/';
```
dstrawberrygirl: Looks like it's safe to remove these lines?

src/code/modules/tutorials.js
kswenson: In general I sometimes left code commented out if I thought that it might be a useful reminder to someone else, e.g. a potentially useful piece of code that isn't currently being called but might potentially be useful at some point. In this case, however, that argument doesn't seem to apply, so I'll remove it.

kswenson:
I'm inclined to leave the dialog as is for now. It's not clear to me how it should change, and Frieda's supposed to be providing the Level 6 dialog which should cover this case. On the technical side, it seems like there should be additional dialog entries associated with each level, just as there is with challenges and missions. The "You've finished the game!" dialog should not be associated with an individual challenge for exactly this reason. I'll take a look at the z-index issue you mentioned.

kswenson:
I just pushed a commit which eliminates the removable code and fixes the z-index issue.

dstrawberrygirl:
Still one small z-index bug when you reach the end of 5.1.3 - the top pair of options pop above the Venture pad (the 'm' non-editable allele style was previously 8 and is now 11). We may want to bring the Venture pad up to allow for the extra options.

kswenson:
@dstrawberrygirl The problem with bumping up the z-index on the VenturePad is that that starts a cascade of other changes, because z-indices of 10, 11, 12, etc. are already in use. Plus, if we ever add more alleles we potentially have to do it again. I still think the right answer is to restrict the z-indices for the allele popups to a local stacking context so that they don't interfere with z-indices outside the genome view. The real problem is that my earlier attempt to create a local stacking context didn't work. I've added another commit which should (really) always create a local stacking context.

src/stylus/fablevision/fv-gene-label.styl
```diff
@@ -1,3 +1,10 @@
+/* establish a stacking context, so all interior z-index values are local */
+.geniblocks.genome
+  transform: scale(1)
```
dstrawberrygirl: Didn't know about this technique before this PR, will definitely use in future!

PR #202, opened 2018-06-06
Enforce minimum number of moves in target-matching challenges

kswenson:
Target-matching challenges now provide a means of specifying a minimum number of moves when generating drakes [#158095906].

### DIFF
--- a/src/code/reducers/helpers/load-state-from-authoring.js
+++ b/src/code/reducers/helpers/load-state-from-authoring.js
@@ -43,6 +43,28 @@ function processAuthoredGametes(authoredChallenge, drakes, state) {
             : state.gametes;
 }
 
+function numberOfMoves(testDrake, targetAlleles, targetSex) {
+  const testSex = testDrake.sex,
+        testAlleles = testDrake.getAlleleString(),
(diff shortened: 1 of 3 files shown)


src/code/reducers/helpers/load-state-from-authoring.js
```diff
@@ -76,28 +98,45 @@ function processAuthoredDrakes(authoredChallenge, authoredTrialNumber, template,
...
         }
       }
+    
+      /*
+       * Guarantee minimum number of moves
+       */
+      const { linkedGenes: { minChanges }, userChangeableGenes } = authoredChallenge;
+      if ((i > 0) && minChanges && userChangeableGenes) {
+        const userGenes = split(userChangeableGenes);
+        while (numberOfMoves(linkedGeneDrake, alleleString, actualDrakeSex) < minChanges) {
```
dstrawberrygirl: Are we safe from an infinite loop situation? Make sure minChanges is validated to acceptable range just in case of authoring glitch.

kswenson: Good thought! I added the additional constraints to the schema.

PR #205, opened 2018-06-06
Add color labels to drakes on color challenges

dstrawberrygirl:
On top of the initial request I also added alt text to all drake images since we had the data readily-available to be added in a human-friendly form in case a drake image fails to load for some reason.

### DIFF
--- a/src/resources/authoring/gv-1.json
+++ b/src/resources/authoring/gv-1.json
@@ -2684,7 +2684,6 @@
     },
     "allele-targetMatch-visible-simpleColors" : {
       "challengeType" : "match-target",
-      "showDrakeColorHint": true,
       "container" : "FVContainer",
       "initialDrake" : [ {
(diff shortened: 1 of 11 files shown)


src/resources/authoring/gv-1.json
```diff
@@ -2684,6 +2684,7 @@
     },
     "allele-targetMatch-visible-simpleColors" : {
       "challengeType" : "match-target",
+      "showDrakeColorHint": true,
```
kswenson: 1. It appears that the authoring document has been auto-formatted to alphabetize properties. I have been maintaining this convention, even though I don't particularly agree with it. (I would prefer dialog sections to be in the order `start, middle, end`, for instance, rather than the reverse.) In this case, putting `showDrakeColorHint` next to the `showUserDrake` property doesn't seem unreasonable.
2. You should add this to the schema for the `FVGenomeChallenge`.

PR #210, opened 2018-06-08
Fix bug which prevented egg game from working correctly when retrying a challenge

kswenson:
Fix bug which prevented FVEggGame challenges from working correctly when retrying a challenge
- onResetGametePools() was being called after generation of the gamete pools but before the challenge

@sfentress fixed [#155271461](https://www.pivotaltracker.com/story/show/155271461) in 0b9583a7baa60d1ba58ad05d316272eb0b229916, which was reverted by @dstrawberrygirl in favor of d8a8f6473de220d85df6357d3e6df1b412e0be14 which was intended to fix [#157810891](https://www.pivotaltracker.com/story/show/157810891), but which also caused [#158224594](https://www.pivotaltracker.com/story/show/158224594).

This should fix all three without reintroducing any of the others.

### DIFF
--- a/src/code/templates/fv-egg-game.js
+++ b/src/code/templates/fv-egg-game.js
@@ -394,7 +394,7 @@ var animationEvents = {
               totalDuration = slotMachineAnimationDuration;
         stage.chroms.forEach(function({ sex, name, sides }) {
           const halfBaseInterval = baseInterval / 2,
-                // initial interval is randomizes so timers aren't synchronized
+                // initial interval is randomized so timers aren't synchronized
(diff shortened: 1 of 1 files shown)


src/code/templates/fv-egg-game.js
```diff
@@ -1287,7 +1287,7 @@ export default class FVEggGame extends Component {
   }
 
   componentWillUnmount() {
-    this.props.onResetGametePools();
+    this.props.onResetGametes();
```
kswenson: This was changed in commit d8a8f6473de220d85df6357d3e6df1b412e0be14 from onResetGametes() to onResetGametePools() without explanation. I'm proposing to change it back.

PR #211, opened 2018-06-08
Tutorials for test cross and sibling pair challenges

dstrawberrygirl:


src/code/modules/tutorial.json
```diff
@@ -71,6 +71,113 @@
...
+        {
+          "mission": 0,
+          "challenges":[
+            {
+              "challenge": 0,
+              "steps":[
+                {
+                  "element": "father",
+                  "text": "Parent Drake",
+                  "more": "This drake’s alleles are invisible!",
```
kswenson: I would change "invisible" to "hidden". The former makes it seems like a property of the drake, as though it's a stealth drake.

src/code/modules/tutorial.json
```diff
@@ -71,6 +71,113 @@
...
+                  "more": "This is the breeder's name for the drake color."
+                }
+              ]
+            }
+          ]
+        }
+      ]
+    },
+    {
+      "level": 3,
```
kswenson: First of two entries for this level.

kswenson:
I couldn't get some of the tutorials to appear. Multiple entries for the same level seem to confuse it.

src/code/modules/tutorial.json
```diff
@@ -71,6 +71,113 @@
...
+                  "more": "Click Breed to produce offspring. Then click a baby drake that matches the target!"
+                }
+              ]
+            }
+          ]
+        }
+      ]
+    },
+    {
+      "level": 3,
```
kswenson: Second entry for this level. The code ignores multiple entries per level.

kswenson:
Two other thoughts not directly related to this PR:
1. We should have a schema for `tutorial.json`
2. It should probably live in the `resources` folder

src/code/modules/tutorial.schema.json
```diff
@@ -0,0 +1,51 @@
...
+                    "properties": {
+                      "steps": {
+                        "description": "The list of tutorial steps for this challenge",
+                        "type": "array",
+                        "items":{
+                          "element": "The css class of the element being highlighted",
+                          "text": "The main header text to show on the popup",
+                          "more": "The full text of the tutorial displayed on clicking More",
+                          "pulseElement": "The element to highlight with a css pulse (if different from the main element",
+                          "location": "A specific location on screen for the popup"
```
kswenson: The items should be in the form:
```
"items": {
  "description": "A tutorial step",
  "type": "object",
  "properties": {
    "element": {
      "description": "The css class of the element being highlighted",
      "type": "string",
    },
    ...
```

dstrawberrygirl: Got it, thanks - figured I'd need a second pass at the schema!

dstrawberrygirl:
The third tutorial of the test cross doesn't have a visible offspring drake to highlight so I have opted to pulse the clutch counter for now. I'll be reaching out for feedback on the best approach here - there is no capacity currently in the tutorial system to show a hint at a different stage of a specific challenge.

PR #213, opened 2018-06-12
Fix target drake gizmo

dstrawberrygirl:
This should get the target gizmo in the correct place for all 4.1.x challenges - Firefox still has other issues with layout that are known and tracked, but at least this tackles one of them.

### DIFF
--- a/src/stylus/fablevision/target-drake.styl
+++ b/src/stylus/fablevision/target-drake.styl
@@ -8,8 +8,6 @@
   background-size: cover
   justify-content: center
   align-self: center
-  margin-bottom: -420px
-  margin-top: -30px
   z-index: 3
 
   .target-drake-text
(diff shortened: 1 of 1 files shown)


src/stylus/fablevision/target-drake.styl
```diff
@@ -1,14 +1,16 @@
...
+  height: 195px
+  width: 500px
   background: url('../resources/fablevision/layout-c/target_drake_gizmo.png')
   background-repeat: no-repeat
   background-size: cover
   justify-content: center
   align-self: center
   z-index: 3
+  position:absolute
+  bottom:0px
```
kswenson: I played around with this in the debugger and with a value of `2px` the gizmo appears to rest on the bottom HUD rather than overlapping it slightly on both Chrome and Firefox. Other than that, this is much simpler than the code it replaces.

PR #215, opened 2018-06-15
Implement challenges 6.4.1 & 6.4.2; initial implementation of 6.4.3

kswenson:
- Add ClutchGame to authoring schema
  - eliminate unused properties of ClutchGame ("instructions", "interactionType", "randomizeTrials", "showUserDrake")
  - fix errors in thirteen invalid allele strings found by validation regex
- Rename challengeType "create-unique" to "match-target" in ClutchGame schema
- GeneticsUtils.convertDashAllelesToABAlleles() issues console errors for invalid allele strings
- Add authoring for mission 6.4 challenges
  - challenge 6.4.1 [#157981366]
  - challenge 6.4.2 [#157981379]
  - challenge 6.4.3 [#157981389]
    - note: challenge 6.4.3 will probably require additional tweaking of drakes
- Fix move counts for challenge 4.2.3
Fix linkedGenes bug in processAuthoredDrakes() which caused all drakes to be linked
- fix authoring for for sibling challenges so that all drakes are linked


kswenson:
Unit test failure is due to unmerged BioLogica changes in another PR.

src/resources/authoring/gv-1.json
```diff
@@ -2195,7 +2195,96 @@
             "text": "Start of mission 6.3"
           }]
         },
-        "name": "Mission 6.3"
+        "name": "Mission 6.4"
```
dstrawberrygirl: This should be "Mission 6.3"

test/utilities/genetics-utils.js
```diff
@@ -27,16 +27,16 @@ describe("GeneticsUtils.convertDashAllelesToABAlleles()", function() {
...
-    assert.equal(abAlleles, "a:W,b:w,a:T,b:Tk,a:dl,b:d,a:Hl,b:hl", "should handle multiple alleles");
+    assert.equal(abAlleles, "a:W,b:w,a:T,b:Tk,a:D,b:d,a:Hl,b:hl", "should handle multiple alleles");
   });
 
-  it("should handle multiple alleles with extraneous spaces", function() {
-    const dashAlleles = " W - w , T - Tk , dl - d , Hl - hl ",
-          abAlleles = GeneticsUtils.convertDashAllelesToABAlleles(dashAlleles);
-    assert.equal(abAlleles, "a:W,b:w,a:T,b:Tk,a:dl,b:d,a:Hl,b:hl", "should handle multiple alleles with spaces");
-  });
+  // it("should handle multiple alleles with extraneous spaces", function() {
```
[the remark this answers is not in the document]
kswenson: Yes. Supporting the additional spaces just fell out of the original implementation as opposed to being an explicit goal. It's not supported by the schema validator, it didn't fall out of the new implementation with error detection, and it's probably a bad idea as it would decrease readability, so I figured it made sense to drop it. All of which seems like an argument for removing it entirely, but I guess I left it in just in case somewhere down the road someone might look at it and have a reason to revive it.

src/resources/authoring/gv-1.json
kswenson: Good catch. Over-aggressive copy-replace. Fixed.

kswenson:
I noticed those things about 6.4.3 as well. I mentioned in the comments above that 6.4.3 will probably require additional work. Perhaps I should have described it more clearly as a work in progress. Several of the issues raised are not particular to 6.4.3, however. The spec document says 6.4.3 should be just like 4.3.2, which also:
- disables the move counter by having zeros for goalMoves. I don't know whether that's an intentional design decision or a bug that should be logged, but it's the current state of sibling challenges.
- does not clear previously bred drakes when alleles are changed. My recollection of earlier discussions was that they should be cleared, but that isn't happening for existing breeding challenges.

On the inability to complete 6.4.3 successfully, the authoring for 6.4.3 mirrors the authoring for 4.3.2 in letting several of the genes be randomly generated and then linked between mother, father, and the targets. I'm not sure where the problem arises but I'm already two days past when I was supposed to stop working on Geniventure, so I probably should turn it over rather than spend more time on it.

kswenson:
Rebased to master after merging the PR with the updated BioLogica dependency. Unit tests are now passing.

dstrawberrygirl:
As @kswenson noted, these challenges match the specification for previous breeding challenges, and the comments I had would reflect on those other existing challenges as much as they would here. I still think there is an issue with 6.4.3 - I was able to breed two parents and see both offspring in the pool, but could not submit them as valid parents. I prematurely approved the PR - I will investigate the problems in 6.4.3 and resolve before proceeding.

PR #218, opened 2018-06-22
Protein game popups

dstrawberrygirl:
Option to re-show protein game popups and placeholder work for scoring / gem color.


dstrawberrygirl:
Added zoom challenge scoring for different gem colors - works on a scale of blue - yellow - red - none for the user performing the correct (vs incorrect) actions 90% (blue) of the time, 80-90% = yellow, 70-80 = red, etc. This will definitely need reviewing based on how easily each challenge can be completed and student results. We can weight the scoring based on authored values if we need to adjust each challenge manually.

src/code/templates/zoom-challenge.js
```diff
@@ -6,10 +6,47 @@ export default class ZoomChallenge extends Component {
...
+            total += stats[m];
+          }
+          // TODO: Calculate actual tolerance for success metric - currently works as:
+          // 90% or higher = blue, 80-90% = yellow, 70-80% = red, etc.
+          const score = positiveTotal / total * 10;
+          const gemScore = 10 - Math.ceil(score);
+          stats.score = gemScore;
+          console.log(stats, score);
+        }
+      }
```
kswenson: My inclination would be to move this block of code into its own function called something `calculateZoomScore()` just to separate the concerns so that `componentDidMount()` is limited to adding the listeners.

src/code/templates/zoom-challenge.js
```diff
@@ -6,10 +6,47 @@ export default class ZoomChallenge extends Component {
...
+
+      // check we have a list of positive metrics and score to calculate
+      if (activityData && activityData.actionStats && positiveMetrics) {
+        stats = activityData.actionStats;
+
+        if (positiveMetrics && stats) {
+          let total = 0, positiveTotal = 0;
+          for (let m in stats) {
+            if (positiveMetrics.indexOf(m) > -1) {
+              positiveTotal += stats[m];
```
kswenson: As we discussed in our zoom meeting, if the metrics were a set of weights rather than a list of strings, then `positiveTotal += stats[m] * (weights[m] || 0)` and `total += stats[m] * Math.abs(weights[m] || 0)` would be a bit more flexible.

src/resources/authoring/gv-1.json
```diff
@@ -2965,12 +2965,22 @@
       "comment" : "*** Level 3, Mission 3.1: Charcoal -> Frost (Albino), Size - Working Everything ***",
       "container" : "FVContainer",
       "template" : "ZoomChallenge",
+      "scoreMetrics": [
+        "disc_clicked",
+        "disc_destroyed",
+        "protein_broken"
+        ],
```
kswenson: As mentioned above, if `scoreMetrics` (or `scoreWeights`) were an object of the form:
```javascript
"scoreWeights": {
  "disc_clicked": 1,
  "disc_destroyed": 1,
  "protein_broken": 1
},
```
then the scoring would be a bit more flexible. In either case, this should ultimately be added to the schema for the zoom challenges (once that schema is written).

PR #223, opened 2018-06-27
Dragon camp - hide hints and fix Firebase issues

dstrawberrygirl:
Commenting out sections of ITS connectivity to ensure the students can play through end-to-end

### DIFF
--- a/src/code/middleware/its-log.js
+++ b/src/code/middleware/its-log.js
@@ -31,19 +31,19 @@ export function initializeITSSocket(guideServer, socketPath, store) {
 
     if (event.isMatch("ITS", "HINT", "USER")) {
       // TODO: Re-enable dispatch once ITS is fixed
-      console.log("ITS provided hint to user:", event.context.dialog);
-      //store.dispatch({type: GUIDE_HINT_RECEIVED, data: event});
+      console.log("ITS provided hint to user:", event);
+      store.dispatch({type: GUIDE_HINT_RECEIVED, data: event});
     } else if (event.isMatch("ITS", "REMEDIATE", "USER")) {
       // TODO: Re-enable dispatch once ITS is fixed
-      console.log("ITS offered user remediation:", event.context.dialog);
-      // store.dispatch(requestRemediation(event));
+      console.log("ITS offered user remediation:", event);
+      store.dispatch(requestRemediation(event));
     } else if (event.isMatch("ITS", "SPOKETO", "USER")) {
       // do nothing with this
       console.log("ITS spoke to user:", event.context.dialog);
     } else if (event.isMatch("ITS", "ISSUED", "ALERT")) {
       // TODO: Re-enable dispatch once ITS is fixed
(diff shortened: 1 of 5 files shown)


src/code/middleware/its-log.js
```diff
@@ -31,19 +31,19 @@ export function initializeITSSocket(guideServer, socketPath, store) {
 
     if (event.isMatch("ITS", "HINT", "USER")) {
       // TODO: Re-enable dispatch once ITS is fixed
```
kswenson: These TODO messages aren't quite right at this point, but things are sufficiently fluid that it probably doesn't matter much.

src/code/middleware/state-save.js
```diff
@@ -41,21 +41,43 @@ export default () => store => next => action => {
...
       };
     }
-    firebase.database().ref(userQueryString).update(userDataUpdate);
+    firebase.database().ref(userQueryString).update(userDataUpdate, (error, userDataUpdate) => {
+      if (error) {
+        console.logError("Error updating user state!", userDataUpdate, error);
+        return error;
+      } else {
+        return;
+      }
```
kswenson: I believe the `else` clause can be skipped here.

src/code/middleware/state-save.js
```diff
@@ -41,21 +41,43 @@ export default () => store => next => action => {
...
-        remediationHistory,
+        // remediationHistory,
         stateVersion: currentStateVersion
       };
     }
-    firebase.database().ref(userQueryString).update(userDataUpdate);
+    firebase.database().ref(userQueryString).update(userDataUpdate, (error, userDataUpdate) => {
+      if (error) {
+        console.logError("Error updating user state!", userDataUpdate, error);
+        return error;
```
kswenson: I'm not seeing any code that's dealing with this error return value. If the point is just to log the error, then maybe the returns can be removed for simplification.

src/code/modules/notifications.js
```diff
@@ -23,27 +23,32 @@ const initialState = {
 export default function notifications(state = initialState, action) {
   switch (action.type) {
     case GUIDE_HINT_RECEIVED:
+      // TODO: Remove when ITS is fixed
+      if (window.location.href.indexOf('enableITS') > -1) {
```
kswenson: We have a `urlParams` module that could be used here.

src/code/middleware/state-save.js
dstrawberrygirl: I want to include more thorough understanding of firebase connection state issues and handle failed updates more gracefully. The above code didn't even get hit when the Firebase update failed (due to undefined data being passed in the remediation section) - definitely flagging this in PT for further investigation.

dstrawberrygirl:
Thanks @kswenson for the feedback. Will integrate these into the next PR to re-enable ITS functionality with the ability to turn it off as required. At least this build seems to be sufficiently stable that students can complete the game without blockers.

PR #234, opened 2018-09-20
Make end of challenge dialog clickable so users can move on to next challenge

kswenson:
Make end of challenge dialog clickable so users can move on to next challenge [#160618770]
- increase the z-index of the modal message so that it's on top of the VenturePad overlay

@sfentress It would be worth taking a look at the overall set of z-indices in use to make sure that there's an appropriate global ordering.


kswenson:
Merged after synchronous review with @ekosmin .

PR #235, opened 2019-04-02
Consecutive egg games had  an incorrect move count

ekosmin:
This was due to hatchSoundPlayed not resetting between levels

### DIFF
--- a/src/code/templates/fv-egg-game.js
+++ b/src/code/templates/fv-egg-game.js
@@ -807,6 +807,7 @@ export default class FVEggGame extends Component {
     _this = this;
     challengeDidChange = true;
     chromosomeDisplayStyle = {display: "none"},
+    hatchSoundPlayed = false;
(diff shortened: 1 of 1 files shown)


ekosmin:
@kswenson Agreed, I don't follow all of the logic here. I wanted to make sure you saw my updated version of the commit...I realized a second bug and both are fixed in the new version + described as well as I could.

src/code/templates/fv-egg-game.js
```diff
@@ -880,7 +880,7 @@ export default class FVEggGame extends Component {
     // after the egg has hatched, reset gametes if the user selects another gamete
     if (hatchSoundPlayed) {
       resetAnimationEvents({ showStaticGametes: true, showHatchAnimation: false });
-      this.props.onResetGametes(true);
+      this.props.onResetGametes(false);
```
ekosmin: I could have left in the change to reset `hatchSoundPlayed` but I decided to leave it alone

PR #242, opened 2019-06-05
Clear remediation state on challenge navigation

kswenson:
Fixes [#166484061](https://www.pivotaltracker.com/story/show/166484061):

>Within a remediation challenge, navigating to another challenge (e.g. using VenturePad or directly modifying the URL) without completing the challenge leaves Geniventure in a confused state that prevents any subsequent remediation from occurring.


kswenson:
@sfentress This explains/fixes the `challenge-id-REMEDIATION` bug we were seeing when looking at the console with Trudi.

PR #258, opened 2020-01-22
Breeding and siblings remediation

sfentress:
This adds breeding and sibling remediation.

As usual, the remediation generator is fairly special-cased and wordy. It attempts to generalize as much as possible, but has to account for the following situations:

1. Straight breeding or sibling (submit parents) challenge
2. One changeable parent or two (straight breeding only)
3. Traits or ITS concepts that are not plain dominance/recessive (sometimes special-cased by concept, sometimes by trait)

### DIFF
--- a/src/code/modules/remediation.js
+++ b/src/code/modules/remediation.js
@@ -417,11 +417,14 @@ function getEggSortRemediation(trait, /* practiceCriteria */) {
   };
 }
 
-function getBreedingRemediation(trait, previousState) {
+function getBreedingRemediation(trait, challengeType, previousState) {
   const templateName = "ClutchGame";
 
+  const isSiblings = challengeType === "Siblings";
   let singleParent = true;
-  if (previousState.userChangeableGenes[0] && previousState.userChangeableGenes[0].mother.length > 0 &&
+  if (!isSiblings && previousState.userChangeableGenes[0] &&
+      previousState.userChangeableGenes[0].mother &&
+      previousState.userChangeableGenes[0].mother.length > 0 &&
       previousState.userChangeableGenes[0].father.length > 0) {
     singleParent = false;
   }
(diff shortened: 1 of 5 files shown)


src/code/modules/remediation.js
```diff
@@ -417,12 +417,126 @@ function getEggSortRemediation(trait, /* practiceCriteria */) {
...
+      fatherAlleles = GeneticsUtils.getAllelesForTrait(trait, "heterozygous");
+      child1Alleles = "a:Tk,b:t";
+      child2Alleles = "a:t,b:t";
+    } else {
+      motherAlleles = GeneticsUtils.getAllelesForTrait(trait, "recessive");
+      fatherAlleles = motherAlleles;
+      child1Alleles = GeneticsUtils.getAllelesForTrait(trait, "dominant");
+      child2Alleles = GeneticsUtils.getAllelesForTrait(trait, "recessive");
+    }
+  }
```
kswenson: As your opening comment suggests, it's tempting to try to find a more general/declarative way to represent this, but it's not obvious to me what that looks like and this is easy to follow (and working) so better just to leave it alone.

src/code/modules/remediation.js
```diff
@@ -417,12 +417,126 @@ function getEggSortRemediation(trait, /* practiceCriteria */) {
...
+      previousState.userChangeableGenes[0].mother.length > 0 &&
+      previousState.userChangeableGenes[0].father.length > 0) {
+    singleParent = false;
+  }
+  let baseAlleles = "a:W,b:W,a:M,b:M,a:T,b:T,a:H,b:H,a:Hl,b:Hl,a:Fl,b:Fl,a:A1,b:A1,a:C,b:C,a:Bog,b:Bog,a:Rh,b:Rh";
+  if (trait === "black") {
+    baseAlleles = baseAlleles + ",a:B,b:B,a:D,b:D";
+  } else {
+    baseAlleles = baseAlleles + ",a:b,b:b,a:d,b:d";
+  }
```
kswenson:
```
  baseAlleles += trait === "black" ? ",a:B,b:B,a:D,b:D" : ",a:b,b:b,a:d,b:d";
```
would be more compact, although debatable as to which is clearer.

src/code/modules/remediation.js
```diff
@@ -417,12 +417,126 @@ function getEggSortRemediation(trait, /* practiceCriteria */) {
   };
 }
 
-export function getRemediationChallengeProps(challengeType, trait, practiceCriteria) {
+function getBreedingRemediation(trait, challengeType, practiceCriteria, previousState) {
+  const templateName = "ClutchGame";
+
+  const isSiblings = challengeType === "Siblings";
+  let singleParent = true;
+  if (!isSiblings && previousState.userChangeableGenes[0] &&
```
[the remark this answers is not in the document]
sfentress: I don't believe so without editing our build system -- or, at least, my linter for this project complains. But it's probably good to do an existence check there, you're right.

src/code/modules/remediation.js
sfentress: True. At one point I had more options, and forgot to refactor.

src/code/modules/remediation.js
sfentress: I could see shrinking this down into a table. The specifics of the `if` clauses would make it seem a bit arbitrary, though.

PR #302, opened 2023-03-01
Update logging server and ITS server URLs

pjanik:
[[#184539865]](https://www.pivotaltracker.com/story/show/184539865)

This small PR mostly updates URLs to point to the AWS servers, letting us move away from Heroku.


kswenson (approved):
👍 LGTM

